PR #3, opened 2022-01-11
Add contracts pallet

ebma:
Adds the contracts pallet to the runtime. The [optional contracts RPC runtime](https://docs.substrate.io/how-to-guides/v3/pallet-design/contracts-pallet/#add-api-dependencies) is not included here because of type issues that I was not able to solve yet. This means that one cannot make calls to the node's storage without making a transaction yet.


gonzamontiel:
Let's include contracts RPC pallet before merging this into master.

ebma:
I now also added the contracts RPC to the node and runtime. 
I don't know why the commit history looks so weird, I just rebased the master onto this branch 😕

gonzamontiel:
Closes #20

gonzamontiel:
Since I just upgraded the runtime to v9.17, we might want to add this changes after #19 is merged, wdyt @ebma ?

gonzamontiel (changes requested):
Please, let's upgrade the versions in this branch to polkadot-v9.17, in compliance with #18

ebma:
That makes sense, I will work on it soon.

runtime/src/lib.rs
```diff
@@ -445,6 +460,46 @@ impl pallet_collator_selection::Config for Runtime {
...
+	type Currency = Balances;
+	type Event = Event;
+	type Call = Call;
+	/// The safest default is to allow no calls at all.
+	///
+	/// Runtimes should whitelist dispatchables that are allowed to be called from contracts
+	/// and make sure they are stable. Dispatchables exposed to contracts are not allowed to
+	/// change because that would break already deployed contracts. The `Call` structure itself
+	/// is not allowed to change the indices of existing pallets, too.
+	type CallFilter = Nothing;
```
gonzamontiel, line 490: So here `Nothing` means that no dispatchables can be called from contracts, right?

gonzamontiel (approved):
left a comment. but lgtm

runtime/src/lib.rs
ebma: I guess so. But apparently it works together with BaseCallFilter. I think keeping it at the default makes the most sense for now.

ebma:
Closing this because it's not relevant and outdated

PR #7, opened 2022-01-26
Add script for running the validator and collator nodes on a single command

gonzamontiel:
Since polkadot-launch is not working for our setup yet, this would work as a workaround. 
The script will assume you have the parachain folder and the polkadot folder under the same root. 
It runs the validator nodes if they're not running yet.
It regenerates the spec files accordingly and then 
it re-runs the parachain node with the new changes.


gonzamontiel:
Run with  `sh ./scripts/test-nodes.sh`
Two validators and one collator will run, output will be redirected to log files. The `PID` of the processes will be printed to properly terminate them when needed

gonzamontiel:
> On macOS I have to turn off my firewall in order for the peers to find themselves. 

Regarding the firewall, I'm afraid the only other way for them to discover each other is to parse the Alice Peer Id from the output  of the first command to include it in the Bob validator. We can include this in a next iteration, but I think this version is a good compromise.

> But besides that, @gonzamontiel would it be possible to leave the script running and when terminating it with Ctrl+C or entering something like "exit" it automatically kills the spawned processes? Would be easier than having to manually kill them with the printed PIDs.
Yes, you're right. Already added it.

gonzamontiel:
Closing this PR since we'll move to use https://github.com/pendulum-chain/pendulum-launch

PR #8, opened 2022-01-26
Fix offchain worker

ebma:
Adds a signing key for 'Alice' to the nodes key store. According to the comment the key should actually be inserted with RPC calls to `author_insertKey` but doing it like this should be sufficient for now.

- [x] Fix Stellar withdrawal (by adding `--enable-offchain-indexing TRUE` to the CLI arguments when running the parachain collator)

### DIFF
--- a/pallets/stellar-bridge/src/lib.rs
+++ b/pallets/stellar-bridge/src/lib.rs
@@ -207,7 +207,7 @@ pub mod pallet {
 		//
 		// For instance you can generate extrinsics for the upcoming produced block.
 		fn offchain_worker(_n: T::BlockNumber) {
-			log::info!("Hello from an offchain worker 👋");
+			log::info!("[OCW] Hello from an offchain worker 👋");
 
 			let res = Self::fetch_latest_txs();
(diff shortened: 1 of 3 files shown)


pallets/stellar-bridge/src/lib.rs
```diff
@@ -263,14 +263,14 @@ pub mod pallet {
...
-				"submit_deposit_unsigned_with_signed_payload: ({:?}, {:?}, {:?})",
+				"[OCW] Submit_deposit_unsigned_with_signed_payload: ({:?}, {:?}, {:?})",
 				amount,
 				destination,
 				signed_by
 			);
 
 			let result = T::Currency::deposit(currency_id, &destination, amount);
-			log::info!("{:?}", result);
+			log::info!("[OCW] {:?}", result);
```
gonzamontiel, line 273: I was looking into this call to see if we could include a prefix in a nicer way, but we can't. A more elegant solution could be wrap the call in a logger defined here, but also this is out of the scope of this pr.

gonzamontiel (approved):
Left a comment, but ready to merge imo.

PR #11, opened 2022-02-02
Declare chain extension in pallet_contracts config

ebma:
Fixes the issue of not being able to deploy the pendulum AMM smart contract.
The `BalanceChainExtension` was defined but never used i.e. not declared in the `pallet_contracts` config. Changing l. 888 fixes the issue.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -522,9 +522,9 @@ impl ChainExtension<Runtime> for BalanceChainExtension {
 		log::info!("Call chain extension: {:?}", func_id,);
 
 		trace!(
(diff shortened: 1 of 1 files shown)


runtime/src/lib.rs
```diff
@@ -520,6 +521,12 @@ impl ChainExtension<Runtime> for BalanceChainExtension {
 	{
 		log::info!("Call chain extension: {:?}", func_id,);
 
+		trace!(
```
gonzamontiel, line 524: Is this well indented?

gonzamontiel (approved):

PR #14, opened 2022-02-09
Bump legacy bridge to polkadot v0.9.16

gonzamontiel:


ebma:
The [canvas repository](https://github.com/paritytech/canvas/tree/1c09bad5b037895a952ed3359a2e5d497be23908) helped in the process of upgrading because it uses both cumulus and the contracts pallet on the latest version (which is `polkadot-v0.9.16`). Just leaving this here for future reference.

ebma (approved):

PR #25, opened 2022-05-23
Bump to polkadot v0918

ne-manya:
This PR bumps the polkadot version to v0.9.18.


xiuxiu62:
Builds and tests are passing on my end.  This looks good, I'll work on spacewalk integration from this branch.  I'm not sure why the test-code workflow is checking cargo registry crates, but that seems to be where the check error is coming from.

xiuxiu62 (approved):

ne-manya:
@xiuxiu62 the error is coming from wrong toolchain version. Pendulum compiles with 1.58 nightly, and on build action is set to 1.57 nightly. This should be changed. 
Anyway it seems to compile on both our ends so I will be merging this now because we need it for other issues.

PR #57, opened 2022-08-05
56 update yamux to 0102

gonzamontiel:
Updates the version of `yamux` to` 0.10.2`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11227,9 +11227,9 @@ version = "1.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "97fee6b57c6a41524a810daee9286c02d7752c4253064d0b05472833a438f675"
 dependencies = [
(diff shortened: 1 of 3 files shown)


TorstenStueber (commented):
Ah I see, the changes are really in Cargo.lock. But its is weird, I see that the package twox-hash now uses a downgraded version of cfg-if and rand and the package yamux uses a later version. How come that there are downgrades without any changes in Cargo.toml?

gonzamontiel:
> Ah I see, the changes are really in Cargo.lock. But its is weird, I see that the package twox-hash now uses a downgraded version of cfg-if and rand and the package yamux uses a later version. How come that there are downgrades without any changes in Cargo.toml?

Hey, this is a result of running `cargo update -p yamux`, how cargo manages the internal  dependencies is out of my reach, but yeah it looks kinda extrange.

TorstenStueber (approved):

PR #60, opened 2022-09-08
Add pallet-vesting, pallet-utility and shorten democracy schedules

TorstenStueber:
I had to add the standard pallet-vesting instead of orml-vesting because orml-vesting does not allow for force-vested-transfers.

A force-vested-transfer is a vested transfer initiated through governance. This allows to circumvent the call filters we set up. We had to make the normal vested-transfer subject to a call filter because it would allow to transfer money – this is not permitted as long as the democratization is not complete.

Since orml-vesting has only the normal vested-transfer, also governance would not be able to pass this call filter and we would not be able to pay out the crowdloan rewards in the next phase.

I don’t see problems to use pallet-vesting instead of orml-vesting for Amplitude

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -124,6 +124,8 @@ dependencies = [
  "pallet-timestamp",
  "pallet-transaction-payment",
  "pallet-transaction-payment-rpc-runtime-api",
+ "pallet-utility",
+ "pallet-vesting",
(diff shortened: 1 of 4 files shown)


b-yap:
substrate-wise, lgtm.

b-yap (approved):

PR #70, opened 2022-10-25
XCM relay native currency  reserve transfer functionality to Pendulum/Amplitude and back 

web3online:


runtime/amplitude/src/xcm_config.rs
```diff
@@ -213,7 +280,133 @@ impl pallet_xcm::Config for Runtime {
...
+            }),
+        }
+    }
+}
+
 impl cumulus_pallet_xcm::Config for Runtime {
 	type Event = Event;
 	type XcmExecutor = XcmExecutor<XcmConfig>;
 }
+
```
TorstenStueber, line 323: Is this still needed? If not, then please remove.

TorstenStueber (changes requested):

PR #81, opened 2022-12-01
Add pallet contract, pallet assets and chain extension to communicate with pallet assets

cr4pt0:


cr4pt0:
mvp version for 0xAmber
this one is deprecated.

we have other implementation with currency pallet. not just pallet asset.

PR #86, opened 2022-12-16
85 create pendulum runtime and chain spec

TorstenStueber:
This PR adds the pendulum runtime and chain spec definitions.

It also adds the initial wasm runtime and genesis state in the `res` folder.


gonzamontiel (approved):
Just one comment that in the future we could extract some behavior to common/, e.g. the contents of `currency.rs`

PR #89, opened 2023-01-06
upgrade to  version v0.9.36 , but it also means updating the toolchain to the latest

b-yap:


TorstenStueber (approved):
@b-yap looks good and makes sense to update the toolchain requirement.

cr4pt0 (approved):

ebma:
@b-yap Is there a reason why you upgraded the dependencies to v0.9.36 and not v0.9.35? I am asking because the spacewalk pallets and testchain are on v0.9.35 and that would probably mean we have to upgrade them to v0.9.36 to make them work with this. Although it's unclear how much effort that would be, could be very easy of course.

node/src/service.rs
```diff
@@ -474,9 +476,7 @@ where
 
 fn build_consensus<RuntimeApi, Executor>(
 	client: Arc<TFullClient<Block, RuntimeApi, NativeElseWasmExecutor<Executor>>>,
-	block_import: TParachainBlockImport<
-		Arc<TFullClient<Block, RuntimeApi, NativeElseWasmExecutor<Executor>>>,
-	>,
+	block_import: ParachainBlockImport<RuntimeApi,Executor>,
```
ebma, line 479: Is the formatting off here?

ebma (commented):
Also looks good to me, but maybe you should run `cargo fmt` again.

node/src/service.rs
b-yap: done

PR #91, opened 2023-01-11
88 add and configure spacewalk pallets for foucoco runtime

b-yap:
todo: after merging https://github.com/pendulum-chain/spacewalk/pull/186 and https://github.com/pendulum-chain/pendulum/pull/89, change branch to `dev/milestone-3`


runtime/foucoco/src/lib.rs
```diff
@@ -706,11 +749,13 @@ impl orml_tokens::Config for Runtime {
 
 parameter_types! {
 	pub const NativeCurrencyId: CurrencyId = CurrencyId::Native;
+	pub const WrappedCurrencyId: CurrencyId = WRAPPED_CURRENCY_ID;
```
ebma, line 752: Do we need this? I don't think we need this anymore. I guess it's actually an orphan statement in the spacewalk testchain I might have forgotten to remove. 🤫

runtime/foucoco/Cargo.toml
```diff
@@ -25,6 +25,27 @@ smallvec = "1.9.0"
...
+replace = { git = "https://github.com/pendulum-chain/spacewalk.git", default-features = false, branch = "update-orml-versions-to-0.9.36" }
+spacewalk-primitives = { git = "https://github.com/pendulum-chain/spacewalk.git", default-features = false, branch = "update-orml-versions-to-0.9.36" }
+
+module-issue-rpc-runtime-api = { git = "https://github.com/pendulum-chain/spacewalk.git", default-features = false, branch = "update-orml-versions-to-0.9.36" }
+module-oracle-rpc-runtime-api = { git = "https://github.com/pendulum-chain/spacewalk.git", default-features = false, branch = "update-orml-versions-to-0.9.36" }
+module-redeem-rpc-runtime-api = { git = "https://github.com/pendulum-chain/spacewalk.git", default-features = false, branch = "update-orml-versions-to-0.9.36" }
+module-replace-rpc-runtime-api = { git = "https://github.com/pendulum-chain/spacewalk.git", default-features = false, branch = "update-orml-versions-to-0.9.36" }
+module-vault-registry-rpc-runtime-api = { git = "https://github.com/pendulum-chain/spacewalk.git", default-features = false, branch = "update-orml-versions-to-0.9.36" }
```
ebma, line 47: Since you removed the `.git` ending in the URLs elsewhere maybe we should remove it here too?

ebma (commented):
From having a very superficial look at it, it seems good so far.

But this actually reminded me that we should look into the `CurrencyId` that we defined in the spacewalk primitives. You configured it correctly I think but maybe we should change that enum overall. Maybe it is not ideal the way it is right now. We should discuss this on slack maybe.

runtime/foucoco/src/lib.rs
b-yap: https://github.com/pendulum-chain/pendulum/pull/91/files#diff-63fad4ac7944e1d379fd236f19986bfb4f8c39eb199a14535775de9a81d72a08R1332

https://github.com/pendulum-chain/pendulum/pull/91/files#diff-f2256017c4e606a30f53708578da46675512cc1300be6542ed5f77e58af0d6dcR581

I think from these

runtime/foucoco/Cargo.toml
b-yap: i was going to put them (`.git`) back today. But I actually don't know which one is better.. having them or not?

runtime/foucoco/src/lib.rs
ebma: Ah okay, interesting. Well, that makes sense actually. But then let's maybe rename it since it merely acts as a default value.
```
--- pub const WrappedCurrencyId: CurrencyId = WRAPPED_CURRENCY_ID;
+++ pub const DefaultWrappedCurrencyId: CurrencyId = WRAPPED_CURRENCY_ID;
```

runtime/foucoco/Cargo.toml
b-yap: i removed the `.git`

runtime/foucoco/src/lib.rs
b-yap: done

cr4pt0 (approved):

b-yap:
this will need @cr4pt0 's https://github.com/pendulum-chain/spacewalk/pull/187 to be pushed, 
and https://github.com/pendulum-chain/spacewalk/pull/194
before this PR will be ready.

cr4pt0 (approved):

node/src/chain_spec.rs
```diff
@@ -365,7 +370,7 @@ pub fn pendulum_config() -> PendulumChainSpec {
 pub fn development_config() -> DevelopmentChainSpec {
 	// Give your base currency a unit name and decimal places
 	let mut properties = sc_chain_spec::Properties::new();
-	properties.insert("tokenSymbol".into(), "UNIT".into());
+	properties.insert("ForeignCurrencyId".into(), "UNIT".into());
```
ebma, line 373: I think this actually has to remain as "tokenSymbol" everywhere. I don't think it is related to our `CurrencyId` enum.

node/src/chain_spec.rs
```diff
@@ -640,11 +704,86 @@ fn foucoco_genesis(
...
+			authorized_oracles,
+			max_delay: 3600000, // one hour
+		},
+		vault_registry: foucoco_runtime::VaultRegistryConfig {
+			minimum_collateral_vault: vec![(XCM(DOT), 0), (XCM(KSM), 0)],
+			punishment_delay: foucoco_runtime::DAYS,
+			secure_collateral_threshold: vec![
+				(default_pair(XCM(DOT)), FixedU128::checked_from_rational(260, 100).unwrap()),
+				(default_pair(XCM(KSM)), FixedU128::checked_from_rational(260, 100).unwrap()),
+			], /* 150% */
```
ebma, line 764: The comments with the values are on the wrong lines, and the percentages are wrong. Should be
260%, 200%, and 150% according to the values used.

node/src/chain_spec.rs
```diff
@@ -217,8 +219,11 @@ pub fn foucoco_config() -> FoucocoChainSpec {
 				// initial collators.
 				invulnerables.clone(),
 				signatories.clone(),
+				//todo: is it okay that only sudo is authorized?
+				vec![(sudo_account.clone(), "Sudo".as_bytes().to_vec())],
```
ebma, line 223: In theory: no. But in practice, we will have to change this anyways once the DIA pallet is implemented. So let's think about it again once that happens.

runtime/foucoco/src/lib.rs
```diff
@@ -188,6 +204,23 @@ const MAXIMUM_BLOCK_WEIGHT: Weight =
...
+// };
+// For Testnet USDC issued by
+const WRAPPED_CURRENCY_ID: CurrencyId = CurrencyId::AlphaNum4 {
+	code: *b"USDC",
+	issuer: [
+		20, 209, 150, 49, 176, 55, 23, 217, 171, 154, 54, 110, 16, 50, 30, 226, 102, 231, 46, 199,
+		108, 171, 97, 144, 240, 161, 51, 109, 72, 34, 159, 139,
+	],
+};
+
```
ebma, line 223: Somehow I am not too happy with this. We can keep it like this for now, but we have to think about how we want to specify all the Stellar assets we want to support and how we do that. 

This is something we should tackle in a follow-up PR. I'll create a ticket for that.

b-yap: i added a TODO comment for this

node/src/chain_spec.rs
b-yap: updated/changed

node/src/chain_spec.rs
```diff
@@ -784,6 +796,15 @@ fn foucoco_genesis(
 			replace_griefing_collateral: FixedU128::checked_from_rational(1, 10).unwrap(), // 10%
 		},
 		nomination: foucoco_runtime::NominationConfig { is_nomination_enabled: false },
+		dia_oracle: foucoco_runtime::DiaOracleConfig {
+			authorized_accounts: authorized_oracles,
+			supported_currencies: vec![foucoco_runtime::AssetId::new(
```
cr4pt0, line 800: specifying only one supported currency at the time that oracle_keys OracleConfig required 3 oracle keys 
- DOT
- USDC
- FeeEstimation

security pallet status will be offline (Oracle Offline error)
Now I figure out that we need hardcode `Key::FeeEstimation` inside DiaOracleAdapter implementation or hardcode 
the case for Key::FeeEstimation and retrieve value from spacewalk oracle storage.
Because there is no way to configure `Key::FeeEstimation` on DIA oracle provider (pallet side)
@ebma @b-yap

ebma: You are right. That's a problem.

cr4pt0: Here is a ticket. 
Will remove Key::FeeEstimation at all.
No one need it.
https://github.com/pendulum-chain/spacewalk/issues/233

cr4pt0 (approved):

cr4pt0 (approved):

ebma (approved):
Code looks good, let me know once you tested if it runs locally and I will approve it @b-yap

b-yap:
closing this.
please go to https://github.com/pendulum-chain/pendulum/compare/88-add-spacewalk-pallets-to-foucoco?expand=1

PR #116, opened 2023-02-13
Add spacewalk pallets to amplitude and expose the spacewalk RPC for the amplitude node

b-yap:
This adds and configures the spacewalk pallets for the amplitude runtime. It also changes the node binary so that the Spacewalk RPC endpoints are exposed properly. 
As configuring correct genesis values for the spacewalk pallets is unnecessary, these do not necessarily reflect how we'd like to have the values on-chain.

Note that the logic for starting 'spacewalk'-nodes for foucoco and amplitude is now ugly code duplicates, but it is too difficult to use a generic definition for what it's worth right now. We should figure this out eventually but due to the urgency, I'd like to keep the duplicated code for now.


node/src/chain_spec.rs
```diff
@@ -564,6 +580,85 @@ fn amplitude_genesis(
...
+		},
+		oracle: amplitude_runtime::OracleConfig {
+			max_delay: u32::MAX,
+			oracle_keys: vec![
+				Key::ExchangeRate(XCM(ForeignCurrencyId::DOT)),
+				Key::ExchangeRate(amplitude_runtime::WRAPPED_CURRENCY_ID),
+			],
+		},
+		vault_registry: amplitude_runtime::VaultRegistryConfig {
+			minimum_collateral_vault: vec![(XCM(DOT), 0), (XCM(KSM), 0)],
```
ebma, line 624: Before we go live on Amplitude we probably want to change these. AFAIK product team is still deciding on the exact values for this.

node/src/chain_spec.rs
```diff
@@ -564,6 +580,85 @@ fn amplitude_genesis(
...
+				amplitude_runtime::StatusCode::Shutdown
+			} else {
+				amplitude_runtime::StatusCode::Error
+			},
+		},
+		oracle: amplitude_runtime::OracleConfig {
+			max_delay: u32::MAX,
+			oracle_keys: vec![
+				Key::ExchangeRate(XCM(ForeignCurrencyId::DOT)),
+				Key::ExchangeRate(amplitude_runtime::WRAPPED_CURRENCY_ID),
```
ebma, line 620: Can we change this to 
```
				Key::ExchangeRate(amplitude_runtime::DefaultWrappedCurrencyId::get()),

```
instead? Boils down to the same thing but I prefer the other function because it mentions that it's the default.

node/src/chain_spec.rs
```diff
@@ -507,6 +518,11 @@ fn amplitude_genesis(
 			),
 	));
 
+	let token_balances = balances
+		.iter()
+		.flat_map(|k| vec![(k.0.clone(), XCM(DOT), 1 << 60), (k.0.clone(), XCM(KSM), 1 << 60)])
```
ebma, line 523: We probably don't want to hand out these tokens for free. So let's set the token balances to something empty for all assets.

node/src/chain_spec.rs
```diff
@@ -564,6 +580,85 @@ fn amplitude_genesis(
 			members: signatories.clone(),
 			..Default::default()
 		},
+		tokens: amplitude_runtime::TokensConfig {
+			// Configure the initial token supply for the native currency and USDC asset
```
ebma, line 584: I'd reword to 
```
			// Configure the initial token supply 

```

node/src/chain_spec.rs
```diff
@@ -564,6 +580,85 @@ fn amplitude_genesis(
...
+			balances: token_balances,
+		},
+		issue: amplitude_runtime::IssueConfig {
+			issue_period: amplitude_runtime::DAYS,
+			issue_minimum_transfer_amount: 1000,
+			limit_volume_amount: None,
+			limit_volume_currency_id: XCM(DOT),
```
ebma, line 591: Just leaving this comment as a reminder that we want to change this.

node/src/chain_spec.rs
b-yap: done

node/src/chain_spec.rs
b-yap: done

node/src/chain_spec.rs
b-yap: done

node/src/chain_spec.rs
```diff
@@ -520,7 +520,7 @@ fn amplitude_genesis(
 
 	let token_balances = balances
 		.iter()
-		.flat_map(|k| vec![(k.0.clone(), XCM(DOT), 1 << 60), (k.0.clone(), XCM(KSM), 1 << 60)])
+		.flat_map(|k| vec![(k.0.clone(), XCM(DOT), 0), (k.0.clone(), XCM(KSM), 0)])
 		.collect();
```
ebma, line 524: Do we even have to iterate over it and assign empty balances? I guess we can just create an empty vec? 😅

node/src/chain_spec.rs
```diff
@@ -166,8 +166,10 @@ pub fn amplitude_config() -> AmplitudeChainSpec {
 				// initial collators.
 				invulnerables.clone(),
 				signatories.clone(),
+				vec![sudo_account.clone()],
```
ebma, line 169: We probably want to consider adding more authorized oracles already in the genesis config. I'll create a separate ticket for generating them but it would need to be added here.

cr4pt0 (approved):

TorstenStueber:
I see that you removed the CurrencyId definition from Amplitude and used one from Spacewalk instead.

Why is that? Spacewalk must be agnostic for what CurrencyId definition the host chain uses. The CurrencyId in Pendulum and Amplitude will definitely be different!

Instead we need to use a mapping from the CurrencyId of the hostchain to a standard CurrencyId type used in Spacewalk (which will resemble Stellar assets).

cr4pt0 (approved):

node/src/chain_spec.rs
ebma: Nevermind, we don't care as genesis is kind of useless

node/src/service.rs
```diff
@@ -239,11 +239,182 @@ async fn build_relay_chain_interface(
 
 // We use this as the RuntimeApi definition in the following function so that we don't need to specify so many generic traits.
 // This might have to change if we want to use this function for other runtimes.
-use foucoco_runtime::RuntimeApi as FoucocoRuntimeApi;
+use amplitude_runtime::RuntimeApi as AmplitudeRuntimeApi;
+
+/// Start a node that exposes the RPC endpoints for the spacewalk pallets for the amplitude runtime.
+#[sc_tracing::logging::prefix_logs_with("Parachain")]
+async fn start_node_impl_spacewalk_amplitude<Executor>(
```
TorstenStueber, line 246: This quite complex function has been duplicated now three times.

The first time it has been duplicated from [`start_node_impl`](https://github.com/pendulum-chain/pendulum/blob/2783abf0a2abe70c3e7258c4aef2941681789fc0/node/src/service.rs#L587) in PR https://github.com/pendulum-chain/pendulum/pull/153 although the only difference was a single line (call `crate::rpc::create_full` vs `crate::rpc::create_full_spacewalk`).

This duplicate did not use the generic parameter `RuntimeApi` so now it needed to be duplicated again.

But this will be painful to maintain correctly with the next Polkadot version updates...

Can we not just use the one function and just add a single Boolean parameter to distinguish between `create_full` and `create_full_spacewalk`?

runtime/amplitude/src/lib.rs
```diff
@@ -933,6 +1036,127 @@ where
...
+}
+parameter_types! {
+	pub const RelayChainCurrencyId: CurrencyId = XCM(0); // 0 is the index of the relay chain in our XCM mapping
+}
+impl currency::Config for Runtime {
+	type UnsignedFixedPoint = UnsignedFixedPoint;
+	type SignedInner = SignedInner;
+	type SignedFixedPoint = SignedFixedPoint;
+	type Balance = Balance;
+	type GetNativeCurrencyId = NativeCurrencyId;
```
TorstenStueber, line 1057: This should be removed, it is not part of the latest change to Spacewalk anymore.

runtime/amplitude/Cargo.toml
```diff
@@ -25,6 +25,26 @@ smallvec = "1.9.0"
 # Local
 runtime-common = {path = "../common", default-features = false}
 
+# Custom libraries for Spacewalk
+currency = { git = "https://github.com/pendulum-chain/spacewalk", default-features = false, branch = "dev/milestone-3" }
```
TorstenStueber, line 29: It is quite risky that we refer to Spacewalk only via a branch name although this branch is still WIP and changing.

Maybe better to add the commit hash (?)

node/src/service.rs
ebma: I updated the [initial description of the issue ticket](https://github.com/pendulum-chain/pendulum/pull/116#issue-1582273849). Did you see that before writing this comment? 

I agree that this is ugly and we should get rid of it. But it seems quite heavy to figure the generic types out and I fear we don't have the time to do that at the moment. If you feel confident that you can figure it out quickly that's great.

runtime/amplitude/src/lib.rs
ebma: Good catch, thanks

runtime/amplitude/Cargo.toml
ebma: I now specified the latest hash of the dev/milestone-3 branch

node/src/service.rs
TorstenStueber: I actually did not read it. So I agree, let's keep it like this right now.

The trait bounds are painful, I agree, and it is one of the pain points of previous Polkadot version updates. Even if I would be able to fix this now, let's not bother with it.

Resolved.

TorstenStueber:
@ebma yes, good point. We are about to open a channel to Bifrost so we might consider to add this. But let's do this at another runtime upgrade.

node/src/service.rs
TorstenStueber: link to issue: https://github.com/pendulum-chain/pendulum/issues/161

TorstenStueber:
I added another commit to bump the runtime version.

cr4pt0 (approved):

PR #126, opened 2023-02-20
Add vesting-manager pallet

TorstenStueber:


runtime/foucoco/src/zenlink.rs
```diff
@@ -7,7 +7,7 @@ use sp_runtime::{DispatchError, DispatchResult};
 use sp_std::marker::PhantomData;
 
 use zenlink_protocol::{
-	AssetId, AssetIdConverter, Config as ZenlinkConfig, LocalAssetHandler, PairLpGenerate,
+	AssetId, AssetIdConverter, LocalAssetHandler, PairLpGenerate,
```
TorstenStueber, line 10: I just removed this reference because it was unused.

pallets/vesting-manager/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
+		VestingScheduleRemoved { who: T::AccountId, schedule_index: u32 },
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::call_index(0)]
+		#[pallet::weight(10_000_000)]
```
TorstenStueber, line 35: I didn't bother to benchmark the actual weight. I just used a high number and then made the transaction feeless if really called from root.

Shall we benchmark it or not worth the time?

ebma: I don't think it's worth the time and we should not add benchmarks for this pallet.

ebma (approved):
LGTM

cr4pt0 (approved):

PR #136, opened 2023-02-28
enable off-chain workers for pendulum node.

cr4pt0:
- enable off-chain workers for pendulum node.
- need redeploy the node. runtime upgrade will not fix issue with workers.


TorstenStueber:
Closed. This is a duplicate of https://github.com/pendulum-chain/pendulum/pull/137

PR #137, opened 2023-02-28
enable off chain workers for pendulum node.

cr4pt0:
- enable off-chain workers for pendulum node.
- need redeploy the node. runtime upgrade will not fix issue with workers.

### DIFF
--- a/runtime/foucoco/src/zenlink.rs
+++ b/runtime/foucoco/src/zenlink.rs
@@ -7,7 +7,7 @@ use sp_runtime::{DispatchError, DispatchResult};
 use sp_std::marker::PhantomData;
 
 use zenlink_protocol::{
-	AssetId, AssetIdConverter, Config as ZenlinkConfig, LocalAssetHandler, PairLpGenerate,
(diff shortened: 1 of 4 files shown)


TorstenStueber (approved):

runtime/foucoco/src/zenlink.rs
```diff
@@ -7,7 +7,7 @@ use sp_runtime::{DispatchError, DispatchResult};
 use sp_std::marker::PhantomData;
 
 use zenlink_protocol::{
-	AssetId, AssetIdConverter, Config as ZenlinkConfig, LocalAssetHandler, PairLpGenerate,
+	AssetId, AssetIdConverter, LocalAssetHandler, PairLpGenerate,
```
gonzamontiel, line 10: is this related to the pr?

gonzamontiel (approved):
I approve it, only left an unimportant comment.

runtime/foucoco/src/zenlink.rs
TorstenStueber: No, but it is an unused import and creates a warning – should have already been removed in a previous commit.

PR #156, opened 2023-03-08
update id

cr4pt0:
Update chain spec for foucoco node


TorstenStueber:
I merged without waiting for CI as this is urgent.

Issue #164, opened 2023-03-13
Expose contract api rpc endpoints for Pendulum/Amplitude/Foucoco


RustNinja:
To upload contract to the chain 
rpc endpoints should be implemented for runtime `impl_runtime_apis` macros

```

impl pallet_contracts::ContractsApi<Block, AccountId, Balance, BlockNumber, Hash>
		for Runtime
	{
		fn call(
			origin: AccountId,
			dest: AccountId,
			value: Balance,
			gas_limit: Option<Weight>,
			storage_deposit_limit: Option<Balance>,
			input_data: Vec<u8>,
		) -> pallet_contracts_primitives::ContractExecResult<Balance> {
			let gas_limit = gas_limit.unwrap_or(RuntimeBlockWeights::get().max_block);
			Contracts::bare_call(
				origin,
				dest,
				value,
				gas_limit,
				storage_deposit_limit,
				input_data,
				CONTRACTS_DEBUG_OUTPUT,
				pallet_contracts::Determinism::Deterministic,
			)
		}

		fn instantiate(
			origin: AccountId,
			value: Balance,
			gas_limit: Option<Weight>,
			storage_deposit_limit: Option<Balance>,
			code: pallet_contracts_primitives::Code<Hash>,
			data: Vec<u8>,
			salt: Vec<u8>,
		) -> pallet_contracts_primitives::ContractInstantiateResult<AccountId, Balance>
		{
			let gas_limit = gas_limit.unwrap_or(RuntimeBlockWeights::get().max_block);
			Contracts::bare_instantiate(
				origin,
				value,
				gas_limit,
				storage_deposit_limit,
				code,
				data,
				salt,
				CONTRACTS_DEBUG_OUTPUT
			)
		}

		fn upload_code(
			origin: AccountId,
			code: Vec<u8>,
			storage_deposit_limit: Option<Balance>,
			determinism: pallet_contracts::Determinism,
		) -> pallet_contracts_primitives::CodeUploadResult<Hash, Balance>
		{
			Contracts::bare_upload_code(origin, code, storage_deposit_limit, determinism)
		}

		fn get_storage(
			address: AccountId,
			key: Vec<u8>,
		) -> pallet_contracts_primitives::GetStorageResult {
			Contracts::get_storage(address, key)
		}
	}

```

PR #165, opened 2023-03-13
164 expose contract api rpc endpoints for runtimes

RustNinja:
- to upload, initiate and call contract need to expose ContractApi for runtime
- add implementation of ContractApi for Pendulum and Foucoco in `impl_runtime_apis`

without this implementation there is no possible to upload the contract to the chain.
issue #164


TorstenStueber:
@RustNinja should we not also add this change to the Amplitude runtime?

RustNinja:
> @RustNinja should we not also add this change to the Amplitude runtime?

@TorstenStueber Oh. sure. you are right. 
Pushed the changes to Amplitude runtime also.

TorstenStueber (approved):

Issue #174, opened 2023-03-17
Add Statemint as Combinations of (Location, Asset) pairs which Pendulum trust as reserves.


RustNinja:
Need to add configuration of Combination of (Location, Asset) to pendulum xcm_config.rs

Reproduce step:
here is extrinsic for [Stetemint](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fsys.ibp.network%2Fstatemint#/extrinsics) to check reserve transfer assets for Statemint USDT to Pendulum chain.

encoded  call data
`0x1f0801010100b92001000101004e5d80a091a712b727ecbb00882bb28d4d64a6e7f6c87654b3243c489623f5100104000002043205011f0002093d000000000000`

Here is the pendulum [block details](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Frpc-pendulum.prd.pendulumchain.tech#/explorer/query/0x92a2b006a30e5bec742a4c072443be78bd6fe6f9f76d6e900e121f838a041cd0) where xcmpQueue pallet emit Fail event with error `UntrustedReserveLocation` 
[error description](https://github.com/paritytech/xcm-format#8-the-types-of-error-in-xcm)

`IsReserve` is configuration for trustable chain as reserve chain for assets. link to [polkadot repo](https://github.com/paritytech/polkadot/blob/master/xcm/xcm-executor/src/config.rs#L42-L43)

Now pendulum support only Native assets in `xcm_executor` config

https://github.com/pendulum-chain/pendulum/blob/main/runtime/pendulum/src/xcm_config.rs#L243

```
pub struct XcmConfig;
impl xcm_executor::Config for XcmConfig {
	...
	type IsReserve = NativeAsset;
	...
}
```

need update configuration to following:

```
/// A `FilterAssetLocation` implementation. Filters multi native assets whose
/// reserve is same with `origin`.
pub struct MultiNativeAsset<ReserveProvider>(PhantomData<ReserveProvider>);
impl<ReserveProvider> FilterAssetLocation for MultiNativeAsset<ReserveProvider>
where
	ReserveProvider: Reserve,
{
	fn filter_asset_location(asset: &MultiAsset, origin: &MultiLocation) -> bool {
		if let Some(ref reserve) = ReserveProvider::reserve(asset) {
			if reserve == origin {
				return true;
			}
		}
		false
	}
}
```

`type IsReserve = MultiNativeAsset<RelativeReserveProvider>;`

PR #181, opened 2023-03-21
update pendulum runtime to support XCM reserve transfer for USDT

RustNinja:
PR related to #174 

- use `MultiNativeAsset<RelativeReserveProvider>` instead of `NativeAsset` for associated type `IsReserve` in `xcm_executor` to transfer token from `Statemint` to `Pendulum`
- use `RelativeReserveProvider` in `orml_xtokens` for `ReserveProvider` associated type to transfer token back.


RustNinja:
Merged by another PR. https://github.com/pendulum-chain/pendulum/pull/192

PR #188, opened 2023-03-29
187 update spacewalk deps and perform foucoco runtime upgrade to fix wrong price calculations

### LINKED ISSUE #187: Update Spacewalk deps and perform Foucoco runtime upgrade to fix wrong price calculations
We should update the spacewalk dependencies to the latest version and deploy a new runtime upgrade on Foucoco so that we can re-do QA.

ebma:
Closes #187.

### DIFF
--- a/runtime/foucoco/src/lib.rs
+++ b/runtime/foucoco/src/lib.rs
@@ -244,7 +244,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("amplitude"),
 	impl_name: create_runtime_str!("amplitude"),
 	authoring_version: 1,
-	spec_version: 8,
+	spec_version: 9,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 20,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1907,7 +1907,7 @@ dependencies = [
 [[package]]
 name = "currency"
 version = "1.2.0"
-source = "git+https://github.com/pendulum-chain/spacewalk?rev=e762e94517d64997ea53e2e272420292b012305b#e762e94517d64997ea53e2e272420292b012305b"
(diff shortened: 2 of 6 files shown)


runtime/foucoco/src/lib.rs
```diff
@@ -244,7 +244,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("amplitude"),
 	impl_name: create_runtime_str!("amplitude"),
 	authoring_version: 1,
-	spec_version: 8,
+	spec_version: 9,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 20,
```
RustNinja, line 250: @ebma 
I believe that we also need to change `transaction_version` once we change the version of spacewalk.
according to this [docs](https://docs.rs/sp-version/latest/sp_version/struct.RuntimeVersion.html)

`This number must change when an existing dispatchable (module ID, dispatch ID) is changed, either through an alteration in its user-level semantics, a parameter added/removed/changed, a dispatchable being removed, a module being removed, or a dispatchable/module changing its index.`

jfy

RustNinja (approved):

runtime/foucoco/src/lib.rs
ebma: But updating the spacewalk dependencies did not change any of the extrinsic calls did it? Or am I missing something? 🤔 The price conversion functions are no dispatchables/extrinsics I guess.

RustNinja: You are absolutely right. It does not change anything to to bump up the transaction_version.
Sorry. @ebma. thought that we changed, but then checked the [diff](https://github.com/pendulum-chain/spacewalk/commit/5183e8730a3a10492d0be80720f68e236e0dee10#diff-7d6a7143358faae325bed7fce8e50c6cc742a6d7947ab2ce351f8ae380da2c9a) and we did not.

b-yap (approved):

Issue #191, opened 2023-03-29
Add integration tests for pendulum runtime


RustNinja:
Issue related to https://github.com/pendulum-chain/pendulum/issues/174

Parity provides the [tool](https://github.com/paritytech/polkadot/tree/release-v0.9.37/xcm/xcm-simulator) for integration tests that able to cover any test cases with parachains, relay chain and XCMP/UMP/DMP communication.

Integration tests will use [polkadot-runtime](https://github.com/paritytech/polkadot/tree/release-v0.9.37/runtime/polkadot/src), `pendulum-runtime` and `statemint-runtime`.

Integration tests based on [xcm-simulator](https://github.com/paritytech/polkadot/blob/release-v0.9.37/xcm/xcm-simulator/src/lib.rs) should cover:
- reserve transfer assets from relay chain to `Pendulum` and back.
- reserve transfer USDT from `Statemint` to Pendulum

PR #192, opened 2023-03-29
Add statemint as trustable chain for pendulum + integration tests

RustNinja:
PR related to https://github.com/pendulum-chain/pendulum/issues/174 and https://github.com/pendulum-chain/pendulum/issues/191

- configured xcm that Statemint is trustable source for reserve transfer assets for Pendulum runtime
- add integration tests and basic configuration for network
    - parachain (pendulum-runtime)
    - polkadot (polkadot-runtime)
    - statemint (statemint-runtime)
- use [xcm-emulator](https://github.com/shaunxw/xcm-simulator/blob/master/xcm-emulator/Cargo.toml) instead of [xcm-simulator](https://github.com/paritytech/polkadot/blob/release-v0.9.37/xcm/xcm-simulator/src/lib.rs)
- add dependency of statemint-runtime from [parity cumulus](https://github.com/paritytech/cumulus).
configure statemint runtime for MockNet.
- register asset with id 1984 to simulate USDT as sufficient on system parachain.
- cover bidirectional transfer from Statemint to Pendulum and back.
- unit test to check transfer DOT from pendulum to relay chain


runtime/integration-tests/pendulum/src/tests.rs
```diff
@@ -0,0 +1,350 @@
...
+
+use polkadot_core_primitives::{AccountId, Balance};
+use polkadot_parachain::primitives::Sibling;
+
+const DOT_FEE: Balance = 3200000000;
+const ASSET_ID: u32 = 1984; //Real USDT Asset ID from Statemint
+const INCORRECT_ASSET_ID: u32 = 0;
+pub const UNIT: Balance = 1_000_000_000_000;
+pub const TEN: Balance = 10_000_000_000_000;
+const FEE: u128 = 421434140;
```
ebma, line 19: Same here, what is this fee value about? Does it have to be this value? Please add a comment above that explains its usage

runtime/integration-tests/pendulum/src/tests.rs
```diff
@@ -0,0 +1,350 @@
...
+
+#[test]
+fn transfer_polkadot_from_relay_chain_to_pendulum() {
```
ebma, line 22: Nit: `transfer_dot_from_relay_chain_to_pendulum()`

runtime/integration-tests/pendulum/src/tests.rs
```diff
@@ -0,0 +1,350 @@
...
+		orml_tokens_before = pendulum_runtime::Tokens::balance(
+			pendulum_runtime::PendulumCurrencyId::XCM(0),
+			&ALICE.into(),
+		);
+	});
+
+	Relay::execute_with(|| {
+		assert_ok!(polkadot_runtime::XcmPallet::reserve_transfer_assets(
+			polkadot_runtime::RuntimeOrigin::signed(ALICE.into()),
+			Box::new(Parachain(2094).into().into()),
```
ebma, line 37: Does it need to be `into().into()` or is it a typo and one `into()` is enough?

ebma: Also, I can see [here](https://github.com/paritytech/polkadot/blob/645723987cf9662244be8faf4e9b63e8b9a1b3a3/xcm/xcm-simulator/example/src/lib.rs#L210) in the example that they wrapped it with the `X1()` junction. Shouldn't we do this as well?

runtime/integration-tests/pendulum/src/tests.rs
```diff
@@ -0,0 +1,350 @@
...
+				pendulum_runtime::PendulumCurrencyId::XCM(0),
+				&ALICE.into()
+			),
+			orml_tokens_before + transfer_amount - DOT_FEE
+		);
+	});
+}
+
+#[test]
+fn transfer_polkadot_from_pendulum_to_relay_chain() {
```
ebma, line 70: Nit: `transfer_dot_from_pendulum_to_relay_chain`

runtime/integration-tests/pendulum/src/tests.rs
```diff
@@ -0,0 +1,350 @@
...
+	MockNet::reset();
+
+	let transfer_dot_amount: Balance = dot(10);
+
+	Relay::execute_with(|| {
+		let after_bob_free_balance = polkadot_runtime::Balances::free_balance(&BOB.into());
+		assert_eq!(after_bob_free_balance, dot(100));
```
ebma, line 77: This should be called `before_bob_free_balance` probably.

ebma: Maybe we can extract the `dot(100)` to an extra variable and put it in line 74. as
```
let expected_base_balance = dot(100)
```

and then you can also replace the `dot(100)` in the assert statement in l. 131 with that variable.

runtime/integration-tests/pendulum/src/tests.rs
```diff
@@ -0,0 +1,350 @@
...
+use xcm_emulator::{Junctions, TestExt};
+
+use polkadot_core_primitives::{AccountId, Balance};
+use polkadot_parachain::primitives::Sibling;
+
+const DOT_FEE: Balance = 3200000000;
+const ASSET_ID: u32 = 1984; //Real USDT Asset ID from Statemint
+const INCORRECT_ASSET_ID: u32 = 0;
+pub const UNIT: Balance = 1_000_000_000_000;
+pub const TEN: Balance = 10_000_000_000_000;
```
ebma, line 18: Maybe rename `TEN` to `TEN_UNITS`

ebma: Or remove it and use `10 * UNIT` in the places it was used instead.

runtime/integration-tests/pendulum/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+#![cfg(test)]
+mod polkadot_test_net;
```
ebma, line 2: Why is the macro annotation just for `polkadot_test_net`?

runtime/integration-tests/pendulum/src/tests.rs
```diff
@@ -0,0 +1,350 @@
...
+	});
+
+	Relay::execute_with(|| {
+		let after_bob_free_balance = polkadot_runtime::Balances::free_balance(&BOB.into());
+		assert_eq!(after_bob_free_balance, dot(100) + transfer_dot_amount - FEE);
+	});
+}
+
+#[test]
+fn statemint_transfer_incorrect_asset_to_pendulum_fails() {
```
ebma, line 136: It's a bit hard to understand what this test is about. Why is the asset `XCM(1)` 'incorrect' and what would we expect? Maybe add some comments to the code.

runtime/integration-tests/pendulum/src/tests.rs
```diff
@@ -0,0 +1,350 @@
...
+			TEN
+		);
+	});
+}
+
+#[test]
+fn statemint_transfer_asset_to_statemint() {
+	statemint_transfer_asset_to_pendulum();
+
+	Statemint::execute_with(|| {});
```
ebma, line 307: Why do we need to have this line? Maybe add a comment

runtime/integration-tests/pendulum/src/tests.rs
```diff
@@ -0,0 +1,350 @@
...
+			),
+			WeightLimit::Limited(10_000_000_000),
+		));
+
+		assert_eq!(
+			TEN - 1 * UNIT, //inital balance - one unit
+			Tokens::balance(PendulumCurrencyId::XCM(1), &AccountId::from(BOB))
+		);
+
+		use pendulum_runtime::{RuntimeEvent, System};
```
ebma, line 336: Maybe move the imports to the top of the function

runtime/integration-tests/pendulum/src/polkadot_test_net.rs
ebma, line 1: Let's maybe rename this file to `network_setup.rs` and combine it with the other `setup.rs` file so that we only have one. I think it's hard to find good naming because they are doing similar things, that's why it might make sense to combine them in the first place.

ebma (changes requested):
Overall this looks good and has some very valuable changes. Great job @RustNinja 👍 

I think I would restructure the folder, so that it is `/runtime/integration-tests/src/pendulum` and not `/runtime/integration-tests/pendulum/src`. Maybe we don't need to have that `pendulum` subfolder anyways.

Also, @RustNinja I assume we would need to change the XCM config for amplitude as well to support non-native assets? Which XCM assets do we actually support with these changes, I suppose it's actually any asset from any chain that we have an open channel to?

RustNinja:
> Overall this looks good and has some very valuable changes. Great job @RustNinja 👍
> 
> I think I would restructure the folder, so that it is `/runtime/integration-tests/src/pendulum` and not `/runtime/integration-tests/pendulum/src`. Maybe we don't need to have that `pendulum` subfolder anyways.
> 
> Also, @RustNinja I assume we would need to change the XCM config for amplitude as well to support non-native assets? Which XCM assets do we actually support with these changes, I suppose it's actually any asset from any chain that we have an open channel to?

Here is the all changes that i made in one [pr](https://github.com/pendulum-chain/pendulum/pull/198)

I would like to keep integration tests separate(in different packages) for each runtime(Pendulum/Amplitude/Foucoco).
The reason is that probably one day we will have a different version of polkadot for different networks. 0.39 for Pendulum and 0.43 for Foucoco. So it will be easy to support integration tests in different packages for each network.

> Also, @RustNinja I assume we would need to change the XCM config for amplitude as well to support non-native assets? Which XCM assets do we actually support with these changes, I suppose it's actually any asset from any chain that we have an open channel to?

@ebma 
by this changes in [xcm_config](https://github.com/pendulum-chain/pendulum/pull/192/files#diff-91848e42611fbf16b8a970242187d57e75083383e231b2a458f8404b2d08e920) Pendulum supports only `USDT` asset_id 1984 from Statemint. 
We can enable them everything but we need a task and approval from product to do it.

Amplitude does not have any [open channel](https://kusama.subscan.io/xcm_dashboard) with other chains. so we do not need to change xcm config for amplitude yet.

runtime/integration-tests/pendulum/src/tests.rs
ebma: Nice, thanks for the great comment. 
Should we move it above the declaration of the test function ie line 135 maybe?

ebma (commented):
@RustNinja so if I understand correctly, USDT (1984) is the only asset that will work because it's the only asset that we implemented a conversion for in the `xcm_config` [here](https://github.com/pendulum-chain/pendulum/blob/1dbf5375e961c0fd52eac67ed13ccccab61de821/runtime/pendulum/src/xcm_config.rs#L61-L70)?

runtime/integration-tests/pendulum/src/tests.rs
ebma: Okay I see. If we don't need it, you can also remove it again. Or also add it to the [AccountId](https://github.com/paritytech/polkadot/blob/645723987cf9662244be8faf4e9b63e8b9a1b3a3/xcm/xcm-simulator/example/src/lib.rs#L211) to make it consistent 😅

runtime/integration-tests/pendulum/src/polkadot_test_net.rs
RustNinja: @ebma 
I found that parallel finance integration tests more well organised
here is the link:
https://github.com/parallel-finance/parallel/tree/master/integration-tests/src

but pretty similar way Acala did also.
https://github.com/AcalaNetwork/Acala/blob/master/runtime/integration-tests/src/lib.rs

runtime/integration-tests/pendulum/src/polkadot_test_net.rs
```diff
@@ -54,9 +54,9 @@ pub fn relay_ext() -> sp_io::TestExternalities {
 	let mut t = frame_system::GenesisConfig::default().build_storage::<Runtime>().unwrap();
 	pallet_balances::GenesisConfig::<Runtime> {
 		balances: vec![
-			(AccountId::from(ALICE), dot(100000)),
-			(AccountId::from(BOB), dot(100)),
-			(para_account_id(2094), 10 * dot(100000)),
+			(AccountId::from(ALICE), one(100000)),
+			(AccountId::from(BOB), one(100)),
+			(para_account_id(2094), 10 * one(100000)),
```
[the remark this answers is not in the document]
RustNinja: renamed to fn units(amount)

runtime/integration-tests/pendulum/src/tests.rs
RustNinja: Done.

RustNinja:
> here
@ebma 

Exactly. We declared [here](https://github.com/pendulum-chain/pendulum/blob/1dbf5375e961c0fd52eac67ed13ccccab61de821/runtime/pendulum/src/xcm_config.rs#L62-L70) that we trust(support) as reserve backed assets only two multilocation: 

`MultiLocation::parent()` for Relay-chain(our parrent)
and 
```
MultiLocation::new(
	1,
	X3(
		Parachain(statemint::PARA_ID),
		PalletInstance(statemint::ASSET_PALLET_ID),
		GeneralIndex(statemint::USDT_ASSET_ID),
	),
))
```
no other assets until we get approval from product team.

runtime/integration-tests/pendulum/src/tests.rs
RustNinja: Done.

ebma:
Now that you added the changes to the Foucoco XCM config as well, maybe we should go ahead and change it also for Amplitude? Would it look any different for Amplitude compared to the other two?

RustNinja:
> Now that you added the changes to the Foucoco XCM config as well, maybe we should go ahead and change it also for Amplitude? Would it look any different for Amplitude compared to the other two?

@ebma 

I added the changes to Foucoco to first upgrade foucoco runtime before Pendulum runtime upgrade and check that it will works with DOT transfer from relay chain to our chain and back.
Anyway it works in integration tests and i just want to use this new xcm settings for Foucoco.
for Amplitude there is no any other reserve backed assets from other chain that we should support so i decide to not touch amplitude runtime without actual requirements for it. 
But you right. There should not be any difference in xcm config for pendulum and amplitude.
So let me know if you fell that i should also change amplitude in this PR.

runtime/pendulum/src/lib.rs
```diff
@@ -14,6 +14,7 @@ use crate::zenlink::*;
 use xcm::v1::MultiLocation;
 use zenlink_protocol::{AssetBalance, MultiAssetsHandler, PairInfo};
 
+pub use currency::CurrencyId as PendulumCurrencyId;
```
TorstenStueber, line 17: This is not used.

TorstenStueber (changes requested):

Cargo.toml
```diff
@@ -11,4 +11,6 @@ members = [
 	"runtime/foucoco",
 	"runtime/pendulum",
 	"runtime/development",
+	"runtime/development",
```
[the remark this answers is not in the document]
RustNinja: Thanks. Done.

TorstenStueber (approved):

runtime/pendulum/src/lib.rs
RustNinja: @TorstenStueber 
but if you want i create a pr just few sec ago to re export only for testing
https://github.com/pendulum-chain/pendulum/pull/200

ebma:
> for Amplitude there is no any other reserve backed assets from other chain that we should support so i decide to not touch amplitude runtime without actual requirements for it.
I see. Well, the reasoning makes sense, I would just be afraid that we miss or forget about it once we need it. But if there really is no need to have it on Amplitude yet then let's not add it there.

PR #196, opened 2023-04-03
Statemint integration tests to cover USDT bidirectional reserve transfer Pendulum <-> Statemint

RustNinja:
PR related to issue https://github.com/pendulum-chain/pendulum/issues/174
- add dependency of statemint-runtime from [parity cumulus](https://github.com/paritytech/cumulus).
- configure statemint runtime for MockNet.
- register asset with id 1984 to simulate USDT as sufficient on system parachain.
- cover bidirectional transfer from Statemint to Pendulum and back.
- unit test to check `transfer DOT from pendulum to relay chain`

### DIFF
--- a/runtime/integration-tests/pendulum/src/lib.rs
+++ b/runtime/integration-tests/pendulum/src/lib.rs
@@ -1,13 +1,13 @@
 use frame_support::{
 	assert_ok,
-	traits::{fungibles::Inspect, GenesisBuild},
+	traits::{fungibles::Inspect, Currency, GenesisBuild},
 };
 use pendulum_runtime::{Balances, PendulumCurrencyId, Runtime, System, Tokens};
 use polkadot_core_primitives::{AccountId, Balance, BlockNumber};
-use polkadot_parachain::primitives::Id as ParaId;
+use polkadot_parachain::primitives::{Id as ParaId, Sibling};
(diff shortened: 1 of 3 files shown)


runtime/integration-tests/pendulum/src/lib.rs
```diff
@@ -43,10 +48,21 @@ decl_test_parachain! {
 	}
 }
 
+decl_test_parachain! {
+	pub struct Statemine {
```
ebma, line 52: This should be called `Statemint` on Polkadot I suppose?

runtime/integration-tests/pendulum/src/lib.rs
```diff
@@ -328,3 +403,149 @@ fn transfer_polkadot_from_pendulum_to_relay_chain() {
...
+				pendulum_runtime::PendulumCurrencyId::XCM(1),
+				&BOB.into()
+			),
+			TEN
+		);
+	});
+}
+
+#[test]
+fn statemine_transfer_asset_to_statemint() {
```
ebma, line 499: This name seems wrong

RustNinja: Oh. Sure.
Statemint!!!
Done

PR #200, opened 2023-04-06
Re export pendulum currency id only for test

RustNinja:
Introduce `testing-utils` feature for pendulum runtime and re export pendulum currency id only for tests


RustNinja:
> @RustNinja can you rebase/merge the latest main branch into this?

It is not necessary already because the root branch already merged to master.
So i can close the pr.

PR #214, opened 2023-04-17
Bump amplitude runtime spec version to 9

TorstenStueber:


TorstenStueber:
This is already 2 months old but has never been reviewed and merged.

ebma (approved):

PR #215, opened 2023-04-17
update dependencies to orml packages

RustNinja:
- add patch.crates-io for orml packages in pendulum


RustNinja:
I created other pr that cover this changes 
https://github.com/pendulum-chain/pendulum/pull/208

Issue #217, opened 2023-04-18
Add Zenlink RPC to pendulum node


RustNinja:
[Zenlink UI](https://dex.zenlink.pro/#/swap) will support interface by default if substrate chain expose Zenlink-protocol endpoinds

Requirements:
-  add dependency to zenlink-protocl-api to pendulum-node
- add trait bound as `ZenlinkProtocolApi` for `create_full_pendulum`, `create_full_amplitude` and `create_full_foucoco` functions inside rpc.rs

vadaynujra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/pendulum/217?planning-poker) with Zenhub @RustNinja @TorstenStueber @ebma

Issue #219, opened 2023-04-19
Update runtimes to `polkadot-v0.9.40` 


ebma:
We should update our runtime dependencies to version `polkadot-v0.9.38` so that we get rid of some dependency conflicts.
For example, we were running into problems because our spacewalk pallets use `jsonrpsee` with version >0.16.0 but other parachain pallets that are on `polkadot-v0.9.37` still use `jsonrpsee` version 0.15.x (e.g. Zenlink) which then leads to conflicts when importing RPC modules from their pallets.

vadaynujra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/pendulum/219?planning-poker) with Zenhub @ashneverdawn @ebma @RustNinja @TorstenStueber

TorstenStueber:
Maybe it makes sense to even switch to Polkadot version 0.9.40 now – it is already supported by orml.

This would also require a follow up task to not just execute the runtime upgrade but first to redeploy all of our collator nodes.

vadaynujra:
Updated version in title to `v0.9.40`

TorstenStueber:
I created a `polkadot-v0.9.40` branch in substrate-stellar-sdk repository so that we are good to go from that side. Probably also need to update dependencies in the other related dependencies:
* https://github.com/pendulum-chain/oracle-pallet
* https://github.com/pendulum-chain/Zenlink-DEX-Module
* https://github.com/pendulum-chain/bifrost

TorstenStueber:
For that reason I want to re-estimate. @vadaynujra is such a feature not available (change an estimate)?

b-yap:
@TorstenStueber  including https://github.com/pendulum-chain/spacewalk, right?

b-yap:
Yes I would think this will be higher than 3. The jump from .37 to .40 is huge. 

A few changes I found: 
* [rename grandpa crates](https://github.com/paritytech/substrate/commit/7c8b89b3a6b603902709fc33fdb10a864b8b4204)
* [rename random-collective-flip crates](https://github.com/paritytech/substrate/pull/13301)
* Weight's `from_ref_time()` is deprecated, so we have to rerun that one again

And there's plenty more. 
Currently on https://github.com/pendulum-chain/oracle-pallet.

TorstenStueber:
[...]

PR #218, opened 2023-04-18
217 add zenlink rpc to pendulum/amplitude/foucoco node

RustNinja:
This PR related to Zenlink AMM integration.
Pendulum node should contains Zenlink protocol RPC endpoints.

PR #220, opened 2023-04-19
AMM RPC endpoints for UI integration

RustNinja:
PR related to https://github.com/pendulum-chain/pendulum/issues/217
and https://github.com/pendulum-chain/Zenlink-DEX-Module/issues/1

PR based on top of this [PR](https://github.com/pendulum-chain/pendulum/pull/208)(it is already merged)

- expose zenlink rpc endpoints from pendulum/amplitude/foucoco runtime


ebma:
@TorstenStueber in case you are wondering why we need _yet another fork_, I looked into this with Nikita and we tried to `patch` it again but it did not work unfortunately. Forking was easiest, and we can get rid of that fork again, once we complete #219 because Zenlink has more up-to-date `jsonrpsee` dependencies on their `polkadot-v0.9.38` branch.

node/src/rpc.rs
```diff
@@ -63,7 +97,7 @@ where
 	let FullDeps { client, pool, deny_unsafe } = deps;
 
 	module.merge(System::new(client.clone(), pool, deny_unsafe).into_rpc())?;
-	module.merge(TransactionPayment::new(client).into_rpc())?;
+	module.merge(TransactionPayment::new(client.clone()).into_rpc())?;
```
ebma, line 100: We don't need to clone it if this is the last time we use it no?

node/src/rpc.rs
```diff
@@ -43,6 +47,36 @@ pub struct FullDeps<C, P> {
...
+	module.merge(System::new(client.clone(), pool, deny_unsafe).into_rpc())?;
+	module.merge(TransactionPayment::new(client.clone()).into_rpc())?;
+	module.merge(ZenlinkProtocol::new(client).into_rpc())?;
+	Ok(module)
+}
+
+/// Instantiate all RPC extensions.
+pub fn create_full_development<C, P>(
+	deps: FullDeps<C, P>,
+) -> Result<RpcExtension, Box<dyn std::error::Error + Send + Sync>>
```
ebma, line 79: Why do we need this now?

node/src/command.rs
```diff
@@ -468,7 +468,7 @@ pub fn run() -> Result<()> {
 						.map_err(Into::into)
 					},
 
-					ChainIdentity::Development => crate::service::start_parachain_node::<development_runtime::RuntimeApi, DevelopmentRuntimeExecutor>(
+					ChainIdentity::Development => crate::service::start_parachain_node_development::<development_runtime::RuntimeApi, DevelopmentRuntimeExecutor>(
```
ebma, line 471: Maybe we should still re-use the `start_parachain_node_pendulum()` for this but pass the `development_runtime::RuntimeApi` to it?

ebma (changes requested):
As I said in the comments I would not introduce a separate development config now. At least not as part of this PR. If this is needed then let's do that somewhere else, after discussing it

node/src/rpc.rs
RustNinja: Done.

node/src/rpc.rs
RustNinja: Because development runtime does not implement API from zenlink.
It means that trait bound on old function does not allow because of
`C::Api: ZenlinkProtocolRuntimeApi<Block, AccountId, AssetId>,`

the solution is to duplicate the function and add trait bound for pendulum/amplitude/foucoco runtime and leave the the prev function for development runtime without changes.

node/src/command.rs
RustNinja: It will not work because development runtime does not implement `Zenlink Runtime Api`.
it means that for now it is not possible to pass development runtime to any function for other runtimes because of the trait bound
`C::Api: ZenlinkProtocolRuntimeApi<Block, AccountId, AssetId>,`

RustNinja:
> As I said in the comments I would not introduce a separate development config now. At least not as part of this PR. If this is needed then let's do that somewhere else, after discussing it

@ebma 
it is not possible to NOT introduce new function for development because of the different implementation between pendulum runtime and development runtime.
Now to expose Zenlink RPC endpoints we need to add new trait bound 
`C::Api: ZenlinkProtocolRuntimeApi<Block, AccountId, AssetId>,`

it means that it not possible anymore to use the same function as for pendulum and development runtime.

ebma (approved):
Okay I see. I was not fully aware what's added to the development runtime. AFAIK we don't really use it. But it's probably the right step to also add other functions for that, now that we basically split everything up per network anyways. 

The rest LGTM.

TorstenStueber:
> @TorstenStueber in case you are wondering why we need _yet another fork_, I looked into this with Nikita and we tried to `patch` it again but it did not work unfortunately. Forking was easiest, and we can get rid of that fork again, once we complete #219 because Zenlink has more up-to-date `jsonrpsee` dependencies on their `polkadot-v0.9.38` branch.

I am a little confused about the issues and PRs and how they are linked. This PR here is linked with an issue (#217), which itself is linked to another PR (#218), then you mention issue #219.

Is there a way to clean up the PR <-> issue structure?

RustNinja:
> > @TorstenStueber in case you are wondering why we need _yet another fork_, I looked into this with Nikita and we tried to `patch` it again but it did not work unfortunately. Forking was easiest, and we can get rid of that fork again, once we complete #219 because Zenlink has more up-to-date `jsonrpsee` dependencies on their `polkadot-v0.9.38` branch.
> 
> I am a little confused about the issues and PRs and how they are linked. This PR here is linked with an issue (#217), which itself is linked to another PR (#218), then you mention issue #219.
> 
> Is there a way to clean up the PR <-> issue structure?

@TorstenStueber 
Done

node/src/service.rs
```diff
@@ -766,6 +771,192 @@ where
 	Ok((task_manager, client))
 }
 
+/// Start a node with the given parachain `Configuration` and relay chain `Configuration`.
```
TorstenStueber, line 774: @RustNinja where is all of this code coming from? Is the code duplication now even getting worse?

If it is just duplicated code, then fine. If there are any crucial changes to the duplicated code here, can you point them out?

Issue #225, opened 2023-04-26
deploy amber contracts to foucoco and test that it works


ashneverdawn:
deploy amber contracts to foucoco and test that it works

amber compiled contracts:
[https://app.zenhub.com/files/428984346/5ba8448e-f9df-4eb1-b9ab-76010834a767/download](https://app.zenhub.com/files/428984346/5ba8448e-f9df-4eb1-b9ab-76010834a767/download)

amber compiled slippage curve:
[https://app.zenhub.com/files/428984346/f18f9044-057a-43f3-b8d8-4f88912403d2/download](https://app.zenhub.com/files/428984346/f18f9044-057a-43f3-b8d8-4f88912403d2/download)

amber deployment scripts:
[https://app.zenhub.com/files/428984346/b5e33e2d-68d1-48ee-a79f-b4454e54427d/download](https://app.zenhub.com/files/428984346/b5e33e2d-68d1-48ee-a79f-b4454e54427d/download)

deployment scripts were written for mumbai, so they cannot be used as is, however the deployment steps can be derived from them.

instead of regular erc20 tokens, we should use wrapped native tokens using our psp22pendulum contract found in our `pendulum-ink-wrapper` repo.

prayagd:
@ashneverdawn Can you please update the status here? Looks like the PR is merged.

ashneverdawn:
yes, PR got merged last week. Today, I'll do the runtime upgrade on Foucoco so I can resume the deployment.

ashneverdawn:
### ----- 06_deploy_swap_pools -----

Deploy Contract: SwapPool
Args: [
mUSD.address,
curve.address,
router.address,
backstop.address,
treasury,
"0xAmber mUSD Swap LP",
"mUSD-LP"
]
Contract Address: 6jkq3j4cq119eF6xZDD2NnohtKJnYEhgXpt8rUJgjJSWC5hY

Deploy Contract: SwapPool
Args: [
mEUR.address,
curve.address,
router.address,
backstop.address,
treasury,
"0xAmber mEUR Swap LP",
"mEUR-LP"
]
Contract Address: 6kS2LNht4gpHck8YwPWGxFADP8x1xRcNpYKg1DLBMNXtEeU8

Deploy Contract: SwapPool
Args: [
mGBP.address,
curve.address,
router.address,
backstop.address,
treasury,
"0xAmber mGBP Swap LP",
"mGBP-LP"
]
[...]

PR #234, opened 2023-05-10
increase contract function parameter limit

ashneverdawn:
increases the contract function parameter limit to 16, to match ethereum's limit.

This is needed because 0xAmber is blocked because it has functions with more parameters than our current limit.
See: https://github.com/pendulum-chain/pendulum/issues/225#issuecomment-1542493083


TorstenStueber (approved):
+1

TorstenStueber:
Strange that the test-code job failed without indicating any reason.

ebma:
I re-ran the failed job and it passed. Ready for merge

ebma (approved):

TorstenStueber (approved):

Issue #230, opened 2023-05-04
Test whether chain extensions and contracts could use CurrencyId directly


ashneverdawn:
Test whether chain extensions and contracts could use CurrencyId directly instead of calling try_get_currency_id_from(type_id: u8, code: [u8;12], issuer: [u8; 32])

1) Does this work and do contracts behave as expected?
2) Does an older already deployed contract continue to point to the same asset whenever CurrencyId is updated?

PR #243, opened 2023-05-31
230 use currencyid in chain extensions

ashneverdawn:
Relating to: https://github.com/pendulum-chain/pendulum/issues/230
updates the chain extensions to take CurrencyId instead of using a mapping to it.

the ink contract is updated to work with CurrencyId in https://github.com/pendulum-chain/pendulum-ink-wrapper/tree/use-currencyid
I already tested that it all works locally. I will create the PR for that one after this one is merged.

### DIFF
--- a/runtime/common/src/chain_ext.rs
+++ b/runtime/common/src/chain_ext.rs
@@ -186,39 +186,8 @@ impl From<dia::CoinInfo> for CoinInfo {
 	}
 }
 
-/// DecodeByteReader is used for reading a Vec<u8>. It implements codec::Input, which is needed by T::decode() to decode scale encoded types.
-pub struct DecodeByteReader {
-	remaining_len: usize,
-	vec: Vec<u8>,
-}
-impl DecodeByteReader {
-	pub fn new(vec: Vec<u8>) -> Self {
-		Self { remaining_len: vec.len(), vec }
-	}
-}
-// Doc comments passed down from the trait explain what its these functions should do, but basically read() reads the exact number of bytes required to fill the given input buffer. remaining_len() is there to track how much is left to read since the buffer may be smaller than the vec, so it might end up being called multiple times until vec is fully read.
-impl Input for DecodeByteReader {
-	fn remaining_len(&mut self) -> Result<Option<usize>, codec::Error> {
-		Ok(Some(self.remaining_len))
-	}
-	fn read(&mut self, into: &mut [u8]) -> Result<(), codec::Error> {
-		let mut vec_index = self.vec.len() - self.remaining_len;
-		for i in 0..into.len() {
-			if vec_index < self.vec.len() {
-				into[i] = self.vec[vec_index];
-				vec_index += 1;
-			} else {
-				into[i] = 0;
-			}
-		}
-		self.remaining_len = self.vec.len() - vec_index;
-		Ok(())
-	}
-}
-
-/// decode uses DecodeByteReader to decode a Vec<u8> into its scale encoded type.
+/// decode gets the slice from a Vec<u8> to decode it into its scale encoded type.
 pub fn decode<T: Decode>(input: Vec<u8>) -> Result<T, codec::Error> {
-	let mut reader = DecodeByteReader::new(input.clone());
-	let value: T = T::decode(&mut reader)?;
-	return Ok(value)
+	let mut input = input.as_slice();
+	T::decode(&mut input)
 }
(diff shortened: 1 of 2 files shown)


runtime/common/src/chain_ext.rs
```diff
@@ -193,4 +184,40 @@ impl From<dia::CoinInfo> for CoinInfo {
...
+impl DecodeByteReader {
+	pub fn new(vec: Vec<u8>) -> Self {
+		Self { remaining_len: vec.len(), vec }
+	}
+}
+impl Input for DecodeByteReader {
+	fn remaining_len(&mut self) -> Result<Option<usize>, codec::Error> {
+		Ok(Some(self.remaining_len))
+	}
+	fn read(&mut self, into: &mut [u8]) -> Result<(), codec::Error> {
```
ebma, line 202: Having a doc comment what exactly this function does would be helpful. If I understand correctly it reads the contents of itself into another array, padding with `0`s if the `into` array is longer than its own content?

runtime/common/src/chain_ext.rs
```diff
@@ -193,4 +184,40 @@ impl From<dia::CoinInfo> for CoinInfo {
...
+	}
+}
+pub fn decode<T: Decode>(input: Vec<u8>) -> Result<T, codec::Error> {
+	let mut reader = DecodeByteReader::new(input.clone());
+	let value: T = T::decode(&mut reader)?;
+	return Ok(value)
+}
+
+pub type Address = [u8; 32];
+pub type Amount = u128;
```
ebma, line 223: I'd move types more to the top

runtime/common/src/chain_ext.rs
```diff
@@ -1,15 +1,27 @@
...
+pub use spacewalk_primitives::{Asset, CurrencyId};
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Encode, Decode, MaxEncodedLen)]
+#[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
 pub enum OriginType {
 	Caller,
 	Address,
 }
+impl From<u8> for OriginType {
+	fn from(origin: u8) -> OriginType {
```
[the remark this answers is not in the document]
ebma: I assume it's arbitrary but maybe we should add a comment that points this out

ebma (commented):
I think I'd prefer having more doc comments explaining how some things work in more detail as it's not always obvious. Overall these are great changes though :+1:

runtime/common/src/chain_ext.rs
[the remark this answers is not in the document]
ashneverdawn: Yeah, it's arbitrary. I added a comment explaining that this is there for the contract to read/write from its own contract memory where it is stored as u8.

runtime/common/src/chain_ext.rs
ashneverdawn: Done.

runtime/common/src/chain_ext.rs
ashneverdawn: Doc comments here are passed down from the Input trait. But I added a regular comment indicating that as well as a short tldr; about what it is for.

TorstenStueber: I agree however, that it is not clear why this struct and trait implementation is here in the first place. Why is it required?

ashneverdawn: It is explained in the comment I just added. T::decode() requires as input a type that implements the codec::Input trait which it uses to read the data it will decode. (In this case the Vec<u8> which is contained in the struct)

TorstenStueber: @ashneverdawn This is still not clear to me: the comment describes the what but not the why. Why would we need to provide our own custom decoder logic and why would it look like this?

ashneverdawn: I thought this was required, because the decode function would not accept the `Vec<u8>` directly. It expected a type that implements this Input trait (which is not implemented for `Vec<u8>`):
https://github.com/paritytech/parity-scale-codec/blob/52c9871cf8c71163afb8845589df63ab6c0e10a6/src/codec.rs#L65-L108
 
But now looking at this again, I see there actually exists an implementation for [u8]
So, I just pushed a commit removing my Vec<u8> implementation and instead calling `.as_slice()` on the `Vec<u8>`

ebma (approved):

TorstenStueber (approved):

PR #242, opened 2023-05-30
upgrade to version `polkadot-v0.9.40`

b-yap:
closes #219

Issue #241, opened 2023-05-26
Update spacewalk dependencies and configure the LP token for use with the Zenlink pallet


ebma:
As part of [this](https://github.com/pendulum-chain/spacewalk/issues/353) ticket, we'll add a type `ZenlinkLPToken()` to the `CurrencyId` enum. 

To configure and use this LP token type with Zenlink, we also need to define some conversion functions between `CurrencyId` and `AssetId`. These can be adopted either from [here](https://github.com/zenlinkpro/Zenlink-DEX-Module/blob/7662acbff71cea3b35db918ceb9370e8929be991/example/primitives/src/lib.rs) or [here](https://github.com/bifrost-finance/bifrost/blob/0eb697cc517daf9103982a6be80d6e7ea8f7e458/node/primitives/src/currency.rs) depending on which `LPToken` type was used in Spacewalk.

With the `TryFrom` implementations in place, we can then get rid of our [zenlink_id_to_currency_id](https://github.com/pendulum-chain/pendulum/blob/44f0d6d46da12810c4cef6b9df3d96620360dc5e/runtime/foucoco/src/zenlink.rs#LL127C4-L127C29) conversion function and replace it with simple calls to `asset_id.try_into()`, see eg [here](https://github.com/bifrost-finance/bifrost/blob/0eb697cc517daf9103982a6be80d6e7ea8f7e458/runtime/bifrost-kusama/src/lib.rs#L1689).

Issue #201, opened 2023-04-06
Update zenlink currency conversion for Amplitude and Pendulum


ebma:
We need to update the '[zenlink_id_to_currency_id](https://github.com/pendulum-chain/pendulum/blob/1dbf5375e961c0fd52eac67ed13ccccab61de821/runtime/foucoco/src/zenlink.rs#L127-L147)' function to something that can also derive our Spacewalk currencies (as we also want to use Spacewalk/Stellar assets on Zenlink).

Now when we define the [conversion from a zenlink to a local currency ID](https://github.com/pendulum-chain/pendulum/blob/1dbf5375e961c0fd52eac67ed13ccccab61de821/runtime/foucoco/src/zenlink.rs#L127-L147), right now we always assume the provided asset_index to be the XCM index. But we somehow have to link to our Stellar assets in this conversion function because we have to be able to derive it from the ZenlinkAssetId.
Since the asset_index is of type `u64` but the XCM index we use in our CurrencyId can only be `u8` I'd say it's safe to start assigning Stellar assets with 256 (because u8 only ranges from 0-255)

```
match asset_id.asset_type {
		NATIVE => Ok(CurrencyId::Native),
		LOCAL => {
			let foreign_id: u8 = asset_id.asset_index.try_into().map_err(|_| {
				log::error!("{} has no Foreign Currency Id match.", asset_id.asset_index);
				()
			})?;

            if foreign_id <= u8::MAX {
			    Ok(XCM(foreign_id))
            } else {
               // TODO add some hard-coded lookup
               match foreign_id {
                    256 => Stellar::StellarNative,
                    257 => yadda yadda ....
                    _ => None
            }
		},
		other => {
			log::error!("Unsupported asset type index:{other}");
			Err(())
		},
	}
```

TorstenStueber:
@ebma makes sense. What currencies shall we encode right now? (also @vadaynujra)

vadaynujra:
The ones in [this](https://github.com/pendulum-chain/spacewalk/issues/311) ticket are enough for now.

vadaynujra:
@ebma is this ticket ready to be worked on and is relevant for the current Zenlink work?

ebma:
Because ticket #241 is probably getting rid of a function mentioned in this ticket, I now marked this ticket as blocked by #241 and we should wait for the outcome of that ticket. This ticket might not be necessary afterward.

adelarja:
[...]

PR #246, opened 2023-06-20
[WIP] 241 update spacewalk dependencies and configure the lp token for use with the zenlink pallet

adelarja:
This PR is related to these issues: [Add ZenlinkLPToken to CurrencyId enum](https://github.com/pendulum-chain/spacewalk/issues/353) and [Add the conversion functions between Zenlink Assets and Spacewalk Assets](https://github.com/pendulum-chain/pendulum/issues/241).

For more details about how the conversion functions work, read [this](https://github.com/pendulum-chain/pendulum/issues/201#issuecomment-1589883661) and [this](https://github.com/pendulum-chain/pendulum/issues/201#issuecomment-1591840337) comment.

At the [beginning](https://github.com/pendulum-chain/pendulum/blob/aaa621d5a735532b01e74f7963ee547654d589ff/runtime/foucoco/src/zenlink.rs#LL172C1-L282C2), the conversion functions were developed using the TryFrom trait. For that, a wrapper struct was needed because neither the TryFrom trait nor the CurrencyId enum was defined in our crate. In the end, we also needed to create a function to handle the wrapper type, so I decided to use functions to convert between Spacewalk assets and Zenlink assets.

To enhance code organization and cleanliness, the conversion functions were divided into [zenlink_id_to_currency_id](https://github.com/pendulum-chain/pendulum/blob/241-update-spacewalk-dependencies-and-configure-the-lp-token-for-use-with-the-zenlink-pallet/runtime/common/src/zenlink.rs#L30) and [currency_id_to_zenlink_id](https://github.com/pendulum-chain/pendulum/blob/241-update-spacewalk-dependencies-and-configure-the-lp-token-for-use-with-the-zenlink-pallet/runtime/common/src/zenlink.rs#L57).

To ensure the code can be used across all our runtimes, it was placed in a separate file within the common directory of the runtime. The currency_id_to_zenlink_id function may not be utilized at the moment, but we can retain it in case it is needed in the future.

Additionally, some unit tests were added to verify the functionality of these functions.

If I'm not mistaken, we'll need [this](https://github.com/pendulum-chain/pendulum/pull/242) pull request (PR) to be merged before we can merge the current one. I made a local test by merging it into my branch, and it successfully resolved several dependency issues between Spacewalk and Pendulum.

### DIFF
--- a/runtime/common/src/zenlink.rs
+++ b/runtime/common/src/zenlink.rs
@@ -28,55 +28,40 @@ fn discriminant(currency: &CurrencyId) -> u8 {
 }
 
 pub fn zenlink_id_to_currency_id(asset_id: ZenlinkAssetId) -> Result<CurrencyId, ()> {
-
 	let _index = asset_id.asset_index;
-		let disc = ((_index & 0x0000_0000_0000_ff00) >> 8) as u8;
-		let symbol = (_index & 0x0000_0000_0000_00ff) as u8;
-		match disc {
-			0 => Ok(CurrencyId::Native),
-			1 => Ok(CurrencyId::XCM(symbol)),
-			2 => match symbol {
-				0 => Ok(CurrencyId::Stellar(Asset::StellarNative)),
-				1 => Ok(CurrencyId::Stellar(Asset::AlphaNum4 {
-					code: *b"USDC",
-					issuer: USDC_ISSUER,
-				})),
-				2 => Ok(CurrencyId::Stellar(Asset::AlphaNum4 {
-					code: *b"TZS\0",
-					issuer: TZS_ISSUER,
-				})),
-				3 => Ok(CurrencyId::Stellar(Asset::AlphaNum4 {
-					code: *b"BRL\0",
-					issuer: BRL_ISSUER,
-				})),
-				_ => return Err(()),
-			},
-			6 => {
-				let token1_id = ((_index & 0x0000_0000_00FF_0000) >> 16) as u8;
-				let token1_type = ((_index & 0x0000_0000_FF00_0000) >> 24) as u8;
+	let disc = ((_index & 0x0000_0000_0000_ff00) >> 8) as u8;
+	let symbol = (_index & 0x0000_0000_0000_00ff) as u8;
+	match disc {
+		0 => Ok(CurrencyId::Native),
+		1 => Ok(CurrencyId::XCM(symbol)),
+		2 => match symbol {
+			0 => Ok(CurrencyId::Stellar(Asset::StellarNative)),
+			1 => Ok(CurrencyId::Stellar(Asset::AlphaNum4 { code: *b"USDC", issuer: USDC_ISSUER })),
+			2 => Ok(CurrencyId::Stellar(Asset::AlphaNum4 { code: *b"TZS\0", issuer: TZS_ISSUER })),
+			3 => Ok(CurrencyId::Stellar(Asset::AlphaNum4 { code: *b"BRL\0", issuer: BRL_ISSUER })),
+			_ => return Err(()),
+		},
+		6 => {
+			let token1_id = ((_index & 0x0000_0000_00FF_0000) >> 16) as u8;
+			let token1_type = ((_index & 0x0000_0000_FF00_0000) >> 24) as u8;
 
-				let token2_id = ((_index & 0x0000_00FF_0000_0000) >> 32) as u8;
-				let token2_type = ((_index & 0x0000_FF00_0000_0000) >> 40) as u8;
+			let token2_id = ((_index & 0x0000_00FF_0000_0000) >> 32) as u8;
+			let token2_type = ((_index & 0x0000_FF00_0000_0000) >> 40) as u8;
 
-				Ok(CurrencyId::ZenlinkLPToken(
-					token1_id,
-					token1_type,
-					token2_id,
-					token2_type,
-				))
-			},
-			_ => Err(()),
-		}
+			Ok(CurrencyId::ZenlinkLPToken(token1_id, token1_type, token2_id, token2_type))
+		},
+		_ => Err(()),
+	}
 }
 
-pub fn currency_id_to_zenlink_id(currency_id: CurrencyId, parachain_id: u32) -> Result<ZenlinkAssetId, ()> {
+pub fn currency_id_to_zenlink_id(
+	currency_id: CurrencyId,
+	parachain_id: u32,
+) -> Result<ZenlinkAssetId, ()> {
 	let disc = discriminant(&currency_id) as u64;
 	match currency_id {
-		CurrencyId::Native => Ok(ZenlinkAssetId {
-			chain_id: parachain_id,
-			asset_type: NATIVE,
-			asset_index: 0 as u64,
-		}),
+		CurrencyId::Native =>
+			Ok(ZenlinkAssetId { chain_id: parachain_id, asset_type: NATIVE, asset_index: 0 as u64 }),
 		CurrencyId::XCM(token_id) => Ok(ZenlinkAssetId {
 			chain_id: parachain_id,
 			asset_type: LOCAL,
@@ -105,15 +90,10 @@ pub fn currency_id_to_zenlink_id(currency_id: CurrencyId, parachain_id: u32) -> 
 				((token1_type as u64) << 24) +
 				((token2_id as u64) << 32) +
 				((token2_type as u64) << 40);
-			Ok(ZenlinkAssetId {
-				chain_id: parachain_id,
-				asset_type: LOCAL,
-				asset_index: _index,
-			})
+			Ok(ZenlinkAssetId { chain_id: parachain_id, asset_type: LOCAL, asset_index: _index })
 		},
 	}
 }
-
 
 #[cfg(test)]
 mod zenlink_tests {
@@ -193,8 +173,7 @@ mod zenlink_tests {
 			asset_index: native_xcm_lp_token_index,
 		};
 		let currency: Result<CurrencyId, _> = zenlink_id_to_currency_id(fake_native_asset);
-		let expected_currency: CurrencyId =
-			CurrencyId::ZenlinkLPToken(0, 0, 0, 1);
+		let expected_currency: CurrencyId = CurrencyId::ZenlinkLPToken(0, 0, 0, 1);
 		assert_eq!(currency, Ok(expected_currency));
 
 		// XCM(0) and XCM(1) LP token Zenlink index = 0x0000_0101_0100_0600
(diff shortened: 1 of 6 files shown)


runtime/common/src/zenlink.rs
```diff
@@ -0,0 +1,257 @@
...
+		},
+		_ => Err(()),
+	}
+}
+
+pub fn currency_id_to_zenlink_id(
+	currency_id: CurrencyId,
+	parachain_id: u32,
+) -> Result<ZenlinkAssetId, ()> {
+	let disc = discriminant(&currency_id) as u64;
```
ebma, line 61: Do we need to cast to `u64` here?

runtime/common/src/zenlink.rs
```diff
@@ -0,0 +1,257 @@
...
+	match currency_id {
+		CurrencyId::Native =>
+			Ok(ZenlinkAssetId { chain_id: parachain_id, asset_type: NATIVE, asset_index: 0 as u64 }),
```
ebma, line 64:
```suggestion
			Ok(ZenlinkAssetId { chain_id: parachain_id, asset_type: NATIVE, asset_index: 0u64 }),
```

runtime/common/src/zenlink.rs
```diff
@@ -0,0 +1,257 @@
...
+			asset_index: ((disc << 8) + token_id as u64) as u64,
+		}),
+		CurrencyId::Stellar(asset) => {
+			let _id = match asset {
+				Asset::StellarNative => 0u64,
+				Asset::AlphaNum4 { code, .. } => match &code {
+					b"USDC" => 1u64,
+					b"TZS\0" => 2u64,
+					b"BRL\0" => 3u64,
+					_ => return Err(()),
```
ebma, line 77: I'd like to check `code` _and_ `issuer` here. It's unlikely going to cause any problems in the future since we are very unlikely going to add two different USDC assets from Stellar. But let's be on the safe side here and make a very explicit mapping that also considers the issuer.

runtime/common/src/zenlink.rs
```diff
@@ -0,0 +1,257 @@
...
+			ZenlinkAssetId { chain_id: 1000u32, asset_type: NATIVE, asset_index: 0 as u64 };
+		let currency: Result<CurrencyId, _> = zenlink_id_to_currency_id(fake_native_asset);
+		assert_eq!(currency, Ok(CurrencyId::Native));
+	}
+
+	#[test]
+	fn convert_zenlink_xcm_to_xcm_currency() {
+		// XCM(0) ZenlinkAsset index = 0x0000_0000_0000_0100
+		let _index = 0x0100 as u64;
+		let fake_native_asset =
```
ebma, line 116: Nit: You probably re-used the name `fake_native_asset` when copy-pasting code for convenience. It doesn't make sense in all functions though.  I'd maybe rename it to `fake_zenlink_asset` in all tests so that it's 'correct' and you don't have to spend time on writing a proper explicit name haha.

runtime/common/src/zenlink.rs
```diff
@@ -0,0 +1,257 @@
...
+fn discriminant(currency: &CurrencyId) -> u8 {
+	match currency {
+		CurrencyId::Native => 0,
+		CurrencyId::XCM(_) => 1,
+		CurrencyId::Stellar(_) => 2,
+		CurrencyId::ZenlinkLPToken(_, _, _, _) => 6,
+	}
+}
+
+pub fn zenlink_id_to_currency_id(asset_id: ZenlinkAssetId) -> Result<CurrencyId, ()> {
```
ebma, line 30: I thought about how we use this function and what it should do.

I realized that it is probably a good idea to check that the `chain_id` contained in the `asset_id` passed to this function matches the ID of our current parachain. At the moment, we do not check the `chain_id` in this function which is problematic.

Let's say pass an `asset_id` for a Bifrost asset here. The `asset_index` conversion logic we implement is roughly the same so we might actually end up treating Bifrost assets like our own. 

Anyways, I think it might be a good idea to add a simple check for the `chain_id` to this function. Or am I exaggerating here and this would never be a problem @adelarja? What do you think?

ebma (commented):
I love it. Thanks for the time and writing the proper tests @adelarja. The tests show quite well how the conversion works. 

I left some comments.

runtime/common/src/zenlink.rs
adelarja: You're right @ebma, I believe it is a great idea.

runtime/common/src/zenlink.rs
adelarja: That's true, my bad haha :joy:.

runtime/common/src/zenlink.rs
adelarja: Since we are using the discriminant to calculate the index (`u64`), we need to cast to `u64`. Maybe we could return a `u64` instead of `u8` in the discriminant function.

runtime/common/src/zenlink.rs
adelarja: @ebma I think it is a good idea.

I'll change the function to receive the parachain_id as a parameter and we can call  `ParachainInfo::parachain_id()` in each parachain runtime configuration when we call the function.

The function signature will change to something like:

```rust
pub fn zenlink_id_to_currency_id(asset_id: ZenlinkAssetId, parachain_id: u32) -> Result<CurrencyId, ()>
```

and we will call the function doing something similar to:

```rust
if let Ok(currency_id) = zenlink_id_to_currency_id(asset_id,  ParachainInfo::parachain_id())
```

What do you think?

ebma: Sounds good 👍

TorstenStueber (commented):
As a general comment: it is super minor but I think we should have standard policy about naming variables. I don't see a need to have some variable names start with an underscore and I don't think that this style is used by the rest of the team.

runtime/common/src/zenlink.rs
```diff
@@ -0,0 +1,272 @@
...
+fn discriminant(currency: &CurrencyId) -> u8 {
+	match currency {
+		CurrencyId::Native => 0,
+		CurrencyId::XCM(_) => 1,
+		CurrencyId::Stellar(_) => 2,
+		CurrencyId::ZenlinkLPToken(_, _, _, _) => 6,
+	}
+}
+
+pub fn zenlink_id_to_currency_id(asset_id: ZenlinkAssetId, parachain_id: u32) -> Result<CurrencyId, ()> {
```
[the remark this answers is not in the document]
adelarja: Yes, that's true. I think it would be a good idea to verify that the asset type is correct. After completing the entire bootstrap process, I realized that we manually set the asset indexes, asset types and chain id. To avoid errors, it would be a good idea to check that we have set the asset type correctly.

runtime/common/src/zenlink.rs
```diff
@@ -0,0 +1,272 @@
+use spacewalk_primitives::{Asset, CurrencyId};
+
+use zenlink_protocol::{LOCAL, NATIVE};
+pub type ZenlinkAssetId = zenlink_protocol::AssetId;
+
+pub const USDC_ISSUER: [u8; 32] = [
```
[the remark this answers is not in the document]
adelarja: Thanks for the comment @TorstenStueber! I put all the issuers information into a separated file.

runtime/common/src/zenlink.rs
```diff
@@ -0,0 +1,267 @@
...
+
+use zenlink_protocol::{LOCAL, NATIVE};
+pub type ZenlinkAssetId = zenlink_protocol::AssetId;
+
+fn discriminant(currency: &CurrencyId) -> u8 {
+	match currency {
+		CurrencyId::Native => 0,
+		CurrencyId::XCM(_) => 1,
+		CurrencyId::Stellar(_) => 2,
+		CurrencyId::ZenlinkLPToken(_, _, _, _) => 6,
```
ebma, line 12: Nit: 
Just something that came to my mind here, we don't actually have to skip 3, 4 and 5, do we? Maybe it's best if we set the discriminant of the LP token to `3`? Or is there any downside to this @adelarja?

adelarja: We need to skip those values because the zenlink protocol uses [this](https://github.com/zenlinkpro/Zenlink-DEX-Module/blob/7662acbff71cea3b35db918ceb9370e8929be991/zenlink-protocol/src/primitives.rs#L65C2-L71C3) function to generate the LP asset id, and the LP token discriminant is hardcoded to 6.

TorstenStueber (approved):

runtime/common/src/zenlink.rs
ebma: Ahhhhh right. Thanks for pointing that out. We should definitely mention that here in a comment though.

Maybe it would be even better if we don't use Zenlink's default implementation of `GenerateLpAssetId` but implement this locally ourselves now? Then we could change it to use the discriminant of `3`. Plus, we wouldn't need to rely on Zenlink to never change that function implementation again. And it would make it a little more obvious how we generate the LP token asset ID. WDYT @adelarja?

ebma: Makes sense. I'd rename `SpacewalkLPGenerate` to `ZenlinkLPGenerate` since this is unrelated to spacewalk. Looks good otherwise 👍

adelarja: @ebma I've made the changes you suggested to me. I have also added the Zenlink Protocol configuration to the Amplitude runtime.

runtime/common/src/stellar.rs
```diff
@@ -0,0 +1,14 @@
+pub const USDC_ISSUER: [u8; 32] = [
```
ebma, line 1:
```suggestion
// GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN
pub const USDC_ISSUER: [u8; 32] = [
```

runtime/common/src/stellar.rs
```diff
@@ -0,0 +1,14 @@
...
+	59, 153, 17, 56, 14, 254, 152, 139, 160, 168, 144, 14, 177, 207, 228, 79, 54, 111, 125, 190,
+	148, 107, 237, 7, 114, 64, 247, 246, 36, 223, 21, 197,
+];
+
+pub const BRL_ISSUER: [u8; 32] = [
```
ebma, line 6:
```suggestion
// GDVKY2GU2DRXWTBEYJJWSFXIGBZV6AZNBVVSUHEPZI54LIS6BA7DVVSP
pub const BRL_ISSUER: [u8; 32] = [
```

runtime/common/src/stellar.rs
```diff
@@ -0,0 +1,14 @@
...
+	234, 172, 104, 212, 208, 227, 123, 76, 36, 194, 83, 105, 22, 232, 48, 115, 95, 3, 45, 13, 107,
+	42, 28, 143, 202, 59, 197, 162, 94, 8, 62, 58,
+];
+
+pub const TZS_ISSUER: [u8; 32] = [
```
ebma, line 11:
```suggestion
// GA2MSSZKJOU6RNL3EJKH3S5TB5CDYTFQFWRYFGUJVIN5I6AOIRTLUHTO 
pub const TZS_ISSUER: [u8; 32] = [
```

ebma (approved):
Everything looks good to me. I just suggested some changes so that we have it a little easier to see what the raw byte keys of the Stellar assets represent. 
I'd say this PR is now ready for merge.

PR #248, opened 2023-06-20
Remove `origin_type` and refactor chain extension **on Foucoco**

### LINKED ISSUE #247: Remove `origin_type` and refactor chain extension

ebma:
This PR removes the `OriginType` field from our chain extension functions, as we don't need it.
It also refactors the code, reordering the functions in accordance with their order in [the ERC20 contract](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/IERC20.sol), and changes the logged warn messages.

I also intended to improve the error handling but apparently, this is not yet possible, because we cannot define custom error types in our runtime, see [this](https://substrate.stackexchange.com/questions/5839/runtimeerror-type) thread.

- [ ] ~~I think I will try to extract the chain-extension code to a separate directory, similar to what Astar did [here](https://github.com/AstarNetwork/Astar/tree/master/chain-extensions). That way we don't have to duplicate code for our runtimes.~~ Follow-up ticket is #249.

Closes #247.

### DIFF
--- a/runtime/foucoco/src/lib.rs
+++ b/runtime/foucoco/src/lib.rs
@@ -991,14 +991,12 @@ where
 					))
 				}
 
-				let result = <orml_currencies::Pallet<T> as MultiCurrency<T::AccountId>>::transfer(
+				<orml_currencies::Pallet<T> as MultiCurrency<T::AccountId>>::transfer(
 					currency_id,
 					&caller,
 					&to,
 					balance,
-				);
-
-				warn!("transfer_result: {:#?}", result);
+				)?;
 			},
 
 			//balance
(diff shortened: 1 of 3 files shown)


ashneverdawn (approved):
lgtm

runtime/foucoco/src/lib.rs
```diff
@@ -962,254 +962,191 @@ where
...
-				}
-
-				let result = orml_currencies_allowance_extension::Pallet::<T>::do_transfer_approved(
-					currency_id,
-					&owner,
-					&from,
-					&to,
-					amount,
+				warn!(
+					"Calling transfer_from() sending {:?} {:?}, from {:?} to {:?}",
```
TorstenStueber, line 1132: Might be reasonable to also add the caller here but not important!

TorstenStueber (approved):

runtime/foucoco/src/lib.rs
```diff
@@ -962,254 +962,191 @@ where
...
-							DispatchError::Other("ChainExtension failed to call 'approve'")
-						})?;
-					},
-				}
+				)?;
 			},
-
-			//transfer_approved
-			1109 => {
+			// transfer_from(currency, sender, recipient, amount)
```
TorstenStueber, line 1116: it is actually sender before currency

ebma: Changed it in [this](https://github.com/pendulum-chain/pendulum/pull/248/commits/8a17e828fef9edf6916d171de319b7e51927f1ad) commit

runtime/foucoco/src/lib.rs
ebma: Done in [this](https://github.com/pendulum-chain/pendulum/pull/248/commits/281c0c7b5baf0770ec90c5a45a1dc74a2ce8f321) commit

ebma:
I decided to also change the function ID of the `get_coin_info()` function that returns price info from the DIA pallet. It seems to me like the `7777` was chosen arbitrarily because it's a nice number. But it probably makes more sense if we start being a bit conservative and assign some kind of number space to certain functionalities. E.g. with the current IDs we could say that `1100-1199` is reserved for functions related to the token pallets and `1200-1299` is reserved for oracle/price data, etc.

TorstenStueber (approved):

PR #272, opened 2023-07-11
Replaced CurrencyId enum by Spacewalk Primitives CurrencyId.

adelarja:
This PR aims to replace the `CurrencyId` enum from pendulum's runtime, with the `CurrencyId` enum from the spacewalk primitives.

We need to do this in order to implement the ZenlinkProtocol Runtime API for pendulum chain in [this](https://github.com/pendulum-chain/pendulum/pull/246) PR.

[zenlink_id_to_currency_id](https://github.com/pendulum-chain/pendulum/blob/271-updatereplace-pendulums-currencyid-with-the-one-defined-in-spacewalk/runtime/pendulum/src/zenlink.rs#L125C1-L128C2) function will be used for the ZenlinkProtocol Runtime API to convert between zenlink assets to spacewalk currencies. The current implementation will be replaced by the same implementation of the [zenlink runtime common module](https://github.com/pendulum-chain/pendulum/blob/b58616fb38a94ab4afa23a659e9dfafeed60f729/runtime/common/src/zenlink.rs#L28C1-L61C2) which was used for Amplitude and Foucoco in the previous PR.


runtime/pendulum/src/zenlink.rs
```diff
@@ -122,41 +122,7 @@ where
...
-				match foreign_currency_id_option {
-					Ok(foreign_currency_id) => Ok(CurrencyId::XCM(foreign_currency_id)),
-					Err(_) => Err(()),
-				}
-			},
-			_ => Err(()),
-		}
-	}
+// This will be replaced with PR 241
+pub fn zenlink_id_to_currency_id(asset_id: ZenlinkAssetId) -> Result<CurrencyId, ()> {
```
adelarja, line 126: zenlink_id_to_currency_id will be replaced with [this](https://github.com/pendulum-chain/pendulum/blob/b58616fb38a94ab4afa23a659e9dfafeed60f729/runtime/common/src/zenlink.rs#L28C1-L61C2) implementation. It might be a good idea to merge the [other PR](https://github.com/pendulum-chain/pendulum/pull/246) first and then merge it into the current branch, allowing me to replace the function.

Alternatively, we can merge it as is, but this function is quite ugly to reach the main branch :joy:

runtime/pendulum/src/lib.rs
```diff
@@ -6,16 +6,15 @@
...
 mod weights;
 pub mod xcm_config;
 pub mod zenlink;
 use crate::zenlink::*;
 use xcm::v3::MultiLocation;
 use zenlink_protocol::{AssetBalance, MultiAssetsHandler, PairInfo};
 
-pub use currency::CurrencyId as PendulumCurrencyId;
 pub use parachain_staking::InflationInfo;
+pub use spacewalk_primitives::CurrencyId as PendulumCurrencyId;
```
ebma, line 17: Nit: I think I would prefer if we don't export this as `PendulumCurrencyId` anymore. I checked the other references and it's only referenced in the integration tests. It's kind of redundant info, since this type is declared/exported in the pendulum runtime. In the related files of the integration tests it's imported as `pendulum_runtime::PendulumCurrencyId` but I think `pendulum_runtime::CurrencyId` would be just as easy to understand.

runtime/pendulum/src/xcm_config.rs
```diff
@@ -1,5 +1,5 @@
 use super::{
-	AccountId, Balance, Balances, CurrencyId, ParachainInfo, ParachainSystem, PolkadotXcm, Runtime,
```
ebma, line 2: If we don't export the `CurrencyId` as `PendulumCurrencyId` as I suggest in the other comment, we can keep this.

runtime/pendulum/src/zenlink.rs
```diff
@@ -6,9 +6,9 @@ use orml_traits::MultiCurrency;
 use sp_runtime::{DispatchError, DispatchResult};
 use sp_std::marker::PhantomData;
 
+use spacewalk_primitives::CurrencyId;
```
ebma, line 9: Also here, if we export the `CurrencyId` without renaming from `lib.rs`, I think we wouldn't even need this `use` statement here because it should be imported by `use super::*` already (?). Should be double-checked

runtime/pendulum/src/zenlink.rs
ebma: I think we can live with this useless dummy implementation for now since we will immediately change it in #241 as you mentioned. Merging this PR first and then integrating the changes to #241 is the better order IMO.

ebma (commented):
I just left some nit-picky comments on how I would prefer the import structure to be. 
I know that the `PendulumCurrencyId` is something that existed before and you just adapted the code to that @adelarja, but while reading through the changes I realized that having that extra type might not be ideal. Again, it's quite minor. Rest LGTM

runtime/pendulum/src/lib.rs
adelarja: Thanks for the comment @ebma

`pendulum_runtime::PendulumCurrencyId` was removed. I've changed the `pendulum_runtime::PendulumCurrencyId` for `pendulum_runtime::CurrencyId` in the integration tests.

runtime/pendulum/src/xcm_config.rs
adelarja: Done!

runtime/pendulum/src/zenlink.rs
adelarja: I've checked this, and it has worked well.

ebma (approved):
Nice, thanks for addressing my comments @adelarja. Looks good to me now 👍

TorstenStueber (approved):
No objections. Looks good.

Issue #178, opened 2023-03-20
Add benchmarks for custom pallets per runtime


ebma:
We should consider benchmarking the custom pallets (spacewalk + orml-extension) for each of our runtimes separately. The default_weights that are included by the pallets were generated on a standalone chain, and the configuration parameters between that test chain and our parachains might differ such that also the resulting weights differ. That's why, for maximum security we should consider re-generating these weights for each of our runtimes. 
This entails:
- [x] Setting up the Foucoco, Amplitude and Pendulum runtime for benchmarking, 
- [x] Running the benchmarks for our custom (spacewalk + orml-extension) pallets to generate the weights, and
- [x] Specifying the generated weights in the respective pallets configurations of the runtimes.

prayagd:
Moving to review

PR #276, opened 2023-07-14
Add and configure asset registry pallet

### LINKED ISSUE #275: Add and configure asset registry pallet
We should add an asset registry pallet to our runtimes so that we can automate the creation of asset mappings for the [global asset registry](https://github.com/colorfulnotion/xcm-global-registry/tree/main).
We should go for the `orml-asset-registry` pallet as it works well with the `orml-tokens` pallet we use.

This is related to [this](https://github.com/pendulum-chain/tasks/issues/75) ticket.

ebma:
Adds the orml-asset-registry pallet to all runtimes. The `CustomMetadata` struct is defined in the spacewalk primitives and should be changed to something that suits our needs. It is still to be unclear what the `CustomMetadata` should be, maybe we just keep it empty. 

## Remaining TODOs
- [x] Change metadata struct in Spacewalk. This has to happen in another repository though and does not require any other code changes here, besides then using the new commit hashes for spacewalk dependencies. ([PR](https://github.com/pendulum-chain/spacewalk/pull/376) pending).

### Notes
The benchmarking for the asset registry was adopted from [here](https://github.com/interlay/interbtc/blob/master/parachain/runtime/common/src/benchmarking/orml_asset_registry.rs). You can find some context on the macros used in the benchmarks [here](https://github.com/paritytech/substrate/pull/12924).

The `AssetAuthority` is configured such that only the root user can register assets. Alternatively we could also give some pallets the permission to register these assets but we don't need this right now.

Closes #275.


runtime/amplitude/src/lib.rs
```diff
@@ -1251,6 +1262,8 @@ construct_runtime!(
 		VaultRewards: reward::{Pallet, Call, Storage, Event<T>} = 70,
 		VaultStaking: staking::{Pallet, Storage, Event<T>} = 71,
 
+		AssetRegistry: orml_asset_registry::{Pallet, Storage, Call, Event<T>, Config<T>} = 91,
```
TorstenStueber, line 1265: Maybe it is good to add a comment here in order not to give the impression that this is another Spacewalk pallet (same for the other `construct_runtime!` invocations in the other runtimes.

TorstenStueber (commented):
I generally approve this PR, particularly how you define the configuration parameters of the asset registry pallet (especially the `AssetProcessor` and the `AuthorityOrigin`).

There is a minor comment I have with the organization of the pallets in `construct_runtime`.

Another thought: this is the first time (?) that we add our own weights for pallets we use. There are some very small differences in the benchmarking results of these pallets for each runtime. I suppose that these are merely statistical deviations and the weights should actually be exactly the same for each runtime.

Would it be more reasonable to move our estimated weights for this pallet into `runtime/common` and avoid all the duplication?

This will set the precedent for further additions of our own benchmarked weights.

ebma:
> Another thought: this is the first time (?) that we add our own weights for pallets we use. There are some very small differences in the benchmarking results of these pallets for each runtime. I suppose that these are merely statistical deviations and the weights should actually be exactly the same for each runtime.

It's actually the second time. As part of the v0.9.40 upgrade, @b-yap recently added custom weights for the XCM pallet, see e.g. [here](https://github.com/pendulum-chain/pendulum/blob/597053d17c24c161ef8729a13e075e78eb74012d/runtime/amplitude/src/weights/pallet_xcm.rs).

> Would it be more reasonable to move our estimated weights for this pallet into runtime/common and avoid all the duplication?

While I see your point with the assumption that the weights should be very similar, I think it would be dangerous to have one set of weights for all runtimes. Generally speaking we of course try to have similar configurations for all our parachains, but they are not perfect copies of each other. It might happen that one day, we change some configuration on Amplitude, that we don't change for Pendulum, which would then lead to different requirements/results for the generated weight.
If you look at other parachain projects, most of them seem to have custom weights _per_ runtime. See:
- [Acala](https://github.com/AcalaNetwork/Acala/tree/master/runtime/acala/src/weights) ✅ 
- Astar ❌ 
- Moonbeam ❌ 
- [interlay](https://github.com/interlay/interbtc/tree/master/parachain/runtime/interlay/src/weights) ✅ 
- [bifrost](https://github.com/bifrost-finance/bifrost/tree/develop/runtime/bifrost-kusama/src/weights) ✅ 
- [mangata](https://github.com/mangata-finance/mangata-node/tree/728fd56be90139db395f1aa8c650d160a7976b67/runtime/mangata-kusama/src/weights) ✅ 
- [centrifuge](https://github.com/centrifuge/centrifuge-chain/tree/main/runtime/centrifuge/src/weights) ✅ 

Astar and Moonbeam seem to use the weights that are defined/exposed by the respective pallets and don't have any custom benchmarked weights.

Thus, I would suggest that we generate the weights for each runtime separately.

I think this topic is a little related to https://github.com/pendulum-chain/pendulum/issues/178 although it does not align perfectly since the ticket is speaking about benchmarks for custom pallets, while we also consider benchmarks for non-custom pallets here.

TorstenStueber:
Okay, sounds reasonable.

TorstenStueber (approved):

runtime/amplitude/src/lib.rs
TorstenStueber: That's fine.

TorstenStueber (approved):

PR #283, opened 2023-07-19
281 inaccuracy in approve chain extension behavior

### LINKED ISSUE #281: Inaccuracy in Approve Chain Extension behavior
While implementing the weight charging for the token allowance pallet, I noticed that the `approve` behavior isn't consistent with OpenZeppelin's one. 

#### 1) The current behavior *increments* the approval, whereas it should *overwrite*. 

See Docs: https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#IERC20-approve-address-uint256-

Current behavior: https://github.com/pendulum-chain/pendulum/blob/0f6cbb5289d03bd85ab664e7724aba691fea8d81/pallets/orml-currencies-allowance-extension/src/lib.rs#L208C3-L208C3

Expected behavior: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/21bb89ef5bfc789b9333eb05e3ba2b7b284ac77c/contracts/token/ERC20/ERC20.sol#L345
Fixing this should also result in a lower gas cost when regenerating the weights. (no longer need to read before writing)

#### 2) When approved for the max possible amount, `transfer_from` should treat it as infinity and not decrement it. 
See Docs: https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#ERC20-approve-address-uint256-

ashneverdawn:
Closes #281 

Corrects `approve` behavior to set rather than increment, and makes `transfer_from` treat u128::MAX as infinite allowance


pallets/orml-currencies-allowance-extension/src/lib.rs
```diff
@@ -298,7 +287,10 @@ impl<T: Config> Pallet<T> {
 				if remaining.is_zero() {
 					*maybe_approved = None;
 				} else {
-					approved = remaining;
+					//decrement allowance only if it isn't max value (which acts as infinite allowance)
+					if approved.clone().checked_into::<u128>().unwrap_or_default() != u128::MAX {
```
ebma, line 291: The `amount` is of type `BalanceOf<T>`. This can be any number type (fulfilling some traits). Let's try not to explicitly use `u128` here. I'd rather try to use the maximum value of the `Balance` type in some way. If I saw correctly, we can use the `max_value()` function on the `Balance` type because it implements the `Bounded` trait.

ebma (commented):
Looks good overall, I requested one improvement.

pallets/orml-currencies-allowance-extension/src/lib.rs
ashneverdawn: Done. Thanks!

PR #284, opened 2023-07-21
257 audit pdm 007 limit size of vector in orml currencies allowance extension pallet

### LINKED ISSUE #257: Audit: PDM-007 - Limit size of vector in `orml-currencies-allowance-extension` pallet
As pointed out by the auditors, we should limit the size of the vector that is used in this pallet. 
Instead of using a regular `Vec` which has an unlimited size, we should instead use a `frame_support::BoundedVec` and define a proper upper limit for that. Otherwise, the extrinsics could be called with excessively large data, potentially leading to long processing times or high resource consumption.

Since the related extrinsics can only be called by the root account, the severity of this issue is rather low.

ebma:
I did not yet manage to make the benchmarks use a variable. 

Closes #257.


pallets/orml-currencies-allowance-extension/src/benchmarking.rs
```diff
@@ -7,7 +7,10 @@ use sp_std::prelude::*;
 benchmarks! {
 	add_allowed_currencies {
```
adelarja, line 8: Not so sure if this helps: I changed the implementation of `add_allowed_currencies` for this one:

```rust
add_allowed_currencies {
		let native_currency_id = <T as orml_currencies::Config>::GetNativeCurrencyId::get();
		let x in 1..T::MaxAllowedCurrencies::get();
	}: add_allowed_currencies(RawOrigin::Root, vec![native_currency_id; x as usize])
	verify {
		let native_currency_id = <T as orml_currencies::Config>::GetNativeCurrencyId::get();
		assert_eq!(AllowedCurrencies::<T>::get(native_currency_id), Some(()));
	}
```
After this change, when running the command indicated in the PR description, I obtained this default_weights.rs file and this output (both files are compressed in the results zip file).

[results.zip](https://github.com/pendulum-chain/pendulum/files/12135702/results.zip)

pallets/orml-currencies-allowance-extension/src/default_weights.rs
```diff
@@ -75,10 +90,61 @@ impl<T: frame_system::Config> crate::WeightInfo for WeightInfo<T> {
...
+			.saturating_add(T::DbWeight::get().reads(4_u64))
+			.saturating_add(T::DbWeight::get().writes(3_u64))
 	}
 }
+
+// For backwards compatibility and tests
+impl WeightInfo for () {
+	/// Storage: TokenAllowance AllowedCurrencies (r:0 w:1)
+	/// Proof Skipped: TokenAllowance AllowedCurrencies (max_values: None, max_size: None, mode: Measured)
+	fn add_allowed_currencies() -> Weight {
```
adelarja, line 104: At the begining, in order for the project to build, I needed to replace this function signature with this one:
```rust
fn add_allowed_currencies(_r: u32) -> Weight
```

pallets/orml-currencies-allowance-extension/src/benchmarking.rs
ebma: Amazing, this helped a lot, thanks 🙏

adelarja (commented):
It seems that the [CI tool is failing](https://github.com/pendulum-chain/pendulum/actions/runs/5644409347/job/15294366467?pr=284#step:9:714) because there is no space on device.

Should we do something similar to what we've done [here](https://github.com/pendulum-chain/spacewalk/blob/main/.github/workflows/ci-main.yml#L22C4-L39C68) on spacewalk?

pallets/orml-currencies-allowance-extension/src/default_weights.rs
```diff
@@ -1,55 +1,77 @@
...
-//! Autogenerated weights for `orml_currencies_allowance_extension`
+//! Autogenerated weights for orml_currencies_allowance_extension
 //!
 //! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
-//! DATE: 2023-07-19, STEPS: `100`, REPEAT: `10`, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! DATE: 2023-07-24, STEPS: `50`, REPEAT: `20`, LOW RANGE: `[]`, HIGH RANGE: `[]`
 //! WORST CASE MAP SIZE: `1000000`
-//! HOSTNAME: `Hugos-MacBook-Pro.local`, CPU: `<UNKNOWN>`
-//! EXECUTION: None, WASM-EXECUTION: Compiled, CHAIN: Some("foucoco"), DB CACHE: 1024
+//! HOSTNAME: `Marcels-MBP`, CPU: `<UNKNOWN>`
```
adelarja, line 7: I'm curious about this. Since the benchmark result depends on the hardware where it runs, would it be a good idea to have a dedicated setup in a VM to run all the benchmarks?

adelarja (approved):

pallets/orml-currencies-allowance-extension/src/default_weights.rs
ebma: That's a very good point. I found [this](https://substrate.stackexchange.com/a/3940/2607) related comment in a forum post, and also [this](https://substrate.stackexchange.com/a/3944/2607) comment (in the same thread) that suggests using a dedicated VM for this purpose, as you suggest.

Running a separate node that meets [these](https://wiki.polkadot.network/docs/maintain-guides-how-to-validate-polkadot#standard-hardware) requirements just for generating the benchmarked weights might be overkill though. However, it's important that we keep this 'problem' in mind. 

@TorstenStueber do you have an opinion on this?

ebma:
@adelarja checks passed now 👌 , can you re-approve?

pallets/orml-currencies-allowance-extension/src/default_weights.rs
TorstenStueber: A defined [another ticket](https://github.com/pendulum-chain/tasks/issues/107) to consider a better solution in the future.

ebma: Perfect, thanks 👍

Issue #83, opened 2022-12-14
Upgrade to Polkadot Version v0.9.33


TorstenStueber:

ebma:
We are on `polkadot-v0.9.37` already so we can close this.

PR #290, opened 2023-08-02
update amplitude runtime to support XCM reserve transfer for USDT

b-yap:
Closes  [task #83](https://github.com/pendulum-chain/tasks/issues/83)

## General overview of the changes:
Aside from the tasks mentioned in the ticket, this PR also  includes:
- adding _Statemine_ and its USDT in the xcm configuration of Amplitude.
- updating the test cases that will cater to both the pendulum and amplitude.
    - moving up the project from `runtime/integration-tests/pendulum` to **`runtime/integration-tests`** 
    - moved ALL _dependencies_ under **_`[dev-dependencies]`_** (see the updated `Cargo.toml`)
    - combined `setup.rs` and `polkadot_test_net.rs` into **`mock.rs`**
    - created a macro for the test cases: **`test_macros.rs`**
    - broke down `tests.rs` to 2: **`pendulum_tests.rs`** and **`amplitude_tests.rs`**

note: Statemint and Statemine have different issued balances, after performing `XTokens::transfer()` in Pendulum and Amplitude.  See line 454 to 456 in **`test_macros.rs`**.


runtime/integration-tests/src/test_macros.rs
```diff
@@ -0,0 +1,468 @@
...
+							asset_id: $para1_asset_id,
+							owner: _,
+							amount,
+						}),
+					topics: _,
+				} => {
+					// https://github.com/paritytech/cumulus/pull/1278 support using self sufficient asset
+					// for paying xcm execution fee on Statemint.
+					// 990_000_000_000 for Statemint
+					// 988_423_297_485 for Statemine
```
b-yap, line 455: this here is what I refer to: 
for Statemint, balance is 990_000_000_000;
for Statemine, balance is 988_423_297_485.

ebma: Why is that?

runtime/common/src/lib.rs
```diff
@@ -76,4 +76,14 @@ pub mod parachains {
 			pub const USDT_ASSET_ID: u128 = 1984;
 		}
 	}
+
+	pub mod kusama {
+		pub mod statemine {
```
ebma, line 81: Let's maybe rename both this `statemine` and the `statemint` above to `asset_hub` as that's the new 'official' name for them.

runtime/integration-tests/src/mock.rs
```diff
@@ -0,0 +1,270 @@
...
+}
+
+pub trait Builder<Currency> {
+	fn balances(self, balances: Vec<(AccountId, Currency, Balance)>) -> Self;
+	fn build(self) -> TestExternalities;
+}
+
+pub enum ParachainType {
+	Statemint,
+	Statemine,
```
ebma, line 107: Also here, I'd just use `AssetHub`.

runtime/integration-tests/src/test_macros.rs
```diff
@@ -0,0 +1,468 @@
...
+			}
+		});
+	}};
+}
+
+pub(super) use parachain1_transfer_asset_to_parachain2;
+pub(super) use parachain1_transfer_asset_to_parachain2_and_back;
+pub(super) use parachain1_transfer_incorrect_asset_to_parachain2_should_fail;
+pub(super) use transfer_10_relay_token_from_parachain_to_relay_chain;
+pub(super) use transfer_20_relay_token_from_relay_chain_to_parachain;
```
ebma, line 468: Should we move this to the top of the file maybe?

runtime/integration-tests/src/mock.rs
b-yap: done

runtime/common/src/lib.rs
b-yap: done

runtime/integration-tests/src/test_macros.rs
b-yap: i tried it before, and it resulted to `unresolved import` and `cannot determine resolution for the macro...`

ebma: Oh, okay 😅 Maybe we should add a little comment then that these are intentionally defined at the bottom of the file and we encounter that error otherwise?

ebma (approved):
Thanks for adding my requested changes @b-yap. 

While at first, it might seem like there is still some code that can be deduplicated, I think it's good the way it is right now. For example, some parameters like the IDs are shared between the Polkadot and Kusama asset hub, but still have separate definitions in `runtime/common/src/lib.rs` but I think it's better this way because we can't necessarily always assume that this is the case. There could have been differences since the Polkadot and Kusama AssetHub are different parachains after all. 

LGTM

runtime/integration-tests/src/test_macros.rs
b-yap: done,

ebma (approved):

ebma:
The previous CI already passed and we only added comments so I will merge it without waiting.

PR #298, opened 2023-08-15
Fix wrong `allowed_currencies_len` variable

### LINKED ISSUE #297: (Amend PDM 007): Fix incorrect allowed-currency counting in currencies allowance pallet

ebma:
Changes the `allowed_currencies_len` to use the length of the updated `AllowedCurrencies` for the second size check (instead of doing a redundant check on the passed `currencies` vector).

Also adds a similar check to the `remove_allowed_currencies` extrinsic. While in theory, it would of course never be possible to remove more currencies than are contained in the `AllowedCurrencies` storage item, which is limited to the `MaxAllowedCurrencies` config parameter in our `add_allowed_currencies` function, the caller could still provide a very large _and valid_ `currencies` vector to this function and thus bloat the chain. That's why also checking the max size of the vector makes sense here.

Closes #297.

### DIFF
--- a/pallets/orml-currencies-allowance-extension/src/lib.rs
+++ b/pallets/orml-currencies-allowance-extension/src/lib.rs
@@ -159,7 +159,7 @@ pub mod pallet {
 
 			// Check if the resulting vector of allowed currencies is less than the maximum allowed.
 			// We check after the insertion to avoid counting duplicates.
-			let allowed_currencies_len: usize = currencies.len();
+			let allowed_currencies_len: usize = AllowedCurrencies::<T>::iter().count();
 			ensure!(
 				allowed_currencies_len <= max_allowed_currencies,
 				Error::<T>::ExceedsNumberOfAllowedCurrencies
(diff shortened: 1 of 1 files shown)


adelarja (approved):
Looks good to me!

TorstenStueber (commented):
Why are the weight definitions of these two extrinsics `add_allowed_currencies(T::MaxAllowedCurrencies::get())` and `remove_allowed_currencies(T::MaxAllowedCurrencies::get())` instead of taking the actual length of the `currencies` argument into account, i.e.:
- `add_allowed_currencies(currencies.len() as u32)` and
- `remove_allowed_currencies(currencies.len() as u32)`

This is what I originally referred to [in this comment](https://github.com/pendulum-chain/pendulum/issues/257#issuecomment-1624703914).

I think that would eliminate all problems.

ebma:
@TorstenStueber as you can see in the [code](https://github.com/paritytech/substrate/blob/5745e3282ce5fccba72538ad03c218124b93b146/frame/identity/src/lib.rs#L395-L397) that you link to in your other comment, when defining the weight in the `[pallet::weight]` macro, they also use the config variable that represents the maximum possible value. If you then look at the actual definition of the `set_sub_olds` weight function declaration [here](https://github.com/paritytech/substrate/blob/5745e3282ce5fccba72538ad03c218124b93b146/frame/identity/src/weights.rs#L429), it's declared using a range variable. This is exactly how we also do it, defining the maximum possible value [here](https://github.com/pendulum-chain/pendulum/blob/21b978be7b794fe254f52a3ded59fa5bc0240f41/pallets/orml-currencies-allowance-extension/src/lib.rs#L141) but using a range variable in the weight function [here](https://github.com/pendulum-chain/pendulum/blob/21b978be7b794fe254f52a3ded59fa5bc0240f41/pallets/orml-currencies-allowance-extension/src/default_weights.rs#L53).

TorstenStueber (approved):
This difference is of course that they also add `T::WeightInfo::set_subs_new(subs.len() as u32)`. But that's really a nitpick. Your solution is fine.

PR #302, opened 2023-08-18
Deduplicate node code

TorstenStueber:
While working on the standalone version to test wasm-deploy, I deduplicated and simplified the code in the `pendulum-node` crate.

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -22,16 +22,16 @@ use crate::constants::{
 
 /// Specialized `ChainSpec` for the normal parachain runtime.
 pub type AmplitudeChainSpec =
-	sc_service::GenericChainSpec<amplitude_runtime::GenesisConfig, Extensions>;
+	sc_service::GenericChainSpec<amplitude_runtime::GenesisConfig, ParachainExtensions>;
 
 pub type FoucocoChainSpec =
-	sc_service::GenericChainSpec<foucoco_runtime::GenesisConfig, Extensions>;
+	sc_service::GenericChainSpec<foucoco_runtime::GenesisConfig, ParachainExtensions>;
 
 pub type PendulumChainSpec =
-	sc_service::GenericChainSpec<pendulum_runtime::GenesisConfig, Extensions>;
+	sc_service::GenericChainSpec<pendulum_runtime::GenesisConfig, ParachainExtensions>;
 
 pub type DevelopmentChainSpec =
-	sc_service::GenericChainSpec<development_runtime::GenesisConfig, Extensions>;
+	sc_service::GenericChainSpec<development_runtime::GenesisConfig, ParachainExtensions>;
 
 /// The default XCM version to set in genesis config.
 const SAFE_XCM_VERSION: u32 = xcm::prelude::XCM_VERSION;
@@ -57,14 +57,14 @@ pub fn create_pendulum_multisig_account(id: &str) -> AccountId {
 /// The extensions for the [`ChainSpec`].
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ChainSpecGroup, ChainSpecExtension)]
 #[serde(deny_unknown_fields)]
-pub struct Extensions {
+pub struct ParachainExtensions {
 	/// The relay chain of the Parachain.
 	pub relay_chain: String,
 	/// The id of the Parachain.
 	pub para_id: u32,
 }
 
-impl Extensions {
+impl ParachainExtensions {
 	/// Try to get the extension from the given `ChainSpec`.
 	pub fn try_get(chain_spec: &dyn sc_service::ChainSpec) -> Option<&Self> {
 		sc_chain_spec::get_extension(chain_spec.extensions())
@@ -160,7 +160,7 @@ pub fn amplitude_config() -> AmplitudeChainSpec {
 		// Properties
 		Some(properties),
 		// Extensions
-		Extensions {
+		ParachainExtensions {
 			relay_chain: "kusama".into(), // You MUST set this to the correct network!
 			para_id: amplitude::PARACHAIN_ID,
 		},
@@ -221,7 +221,7 @@ pub fn foucoco_config() -> FoucocoChainSpec {
 		// Properties
 		Some(properties),
 		// Extensions
-		Extensions {
+		ParachainExtensions {
 			relay_chain: "kusama".into(), // You MUST set this to the correct network!
 			para_id: foucoco::PARACHAIN_ID,
 		},
@@ -344,7 +344,7 @@ pub fn pendulum_config() -> PendulumChainSpec {
 		// Properties
 		Some(properties),
 		// Extensions
-		Extensions {
+		ParachainExtensions {
 			relay_chain: "polkadot".into(), // You MUST set this to the correct network!
 			para_id: pendulum::PARACHAIN_ID,
 		},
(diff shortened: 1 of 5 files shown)


node/src/chain_spec.rs
```diff
@@ -57,14 +57,14 @@ pub fn create_pendulum_multisig_account(id: &str) -> AccountId {
 /// The extensions for the [`ChainSpec`].
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ChainSpecGroup, ChainSpecExtension)]
 #[serde(deny_unknown_fields)]
-pub struct Extensions {
+pub struct ParachainExtensions {
```
TorstenStueber, line 60: I renamed this to distinguish from a standalone chain spec, which does not have this chain spec extension.

node/src/chain_spec.rs
```diff
@@ -394,61 +394,6 @@ pub fn development_config() -> DevelopmentChainSpec {
...
-		None,
-		None,
-		Extensions {
-			relay_chain: "rococo-local".into(), // You MUST set this to the correct network!
-			para_id: 1000,
-		},
-	)
-}
-
-pub fn local_testnet_config() -> DevelopmentChainSpec {
```
TorstenStueber, line 409: It was confusing that we have both a local testnet and a development testnet. I don't think that we need this distinction as both chain specs are basically the same. I only kept the testnet definition.

node/src/command.rs
```diff
@@ -31,21 +32,68 @@ enum ChainIdentity {
 	Development,
 }
 
+impl ChainIdentity {
```
TorstenStueber, line 35: I think that the code is cleaner by adding some specific `ChainIdentity` based logic and `match` expressions to the `ChainIdentity` enum itself. This way some of the `match` expressions in the rest of the code can be removed.

Furthermore I unified the code in the previous `identify` and `load_spec` function which was strangely different: whereas one tested for string equality, the other one tested only whether initial fragments of string are equal. Testing for equality is enough as the only `id`s we deal with are the ones created in the file `chain_spec` or that already exist in any of our precomputed chain spec files in the folder `res`.

node/src/command.rs
```diff
@@ -55,28 +103,6 @@ impl<T: sc_service::ChainSpec + 'static> IdentifyChain for T {
 	}
 }
 
-fn load_spec(id: &str) -> std::result::Result<Box<dyn ChainSpec>, String> {
```
TorstenStueber, line 58: I broke this function down into the `create_chain_spec` and `load_chain_spec_from_json_file` functions of `ChainIdentity`. This simplified this function so that I could directly move the simplified content to [where it is used](https://github.com/pendulum-chain/pendulum/pull/302/files#diff-8ce03a5a2b79f3021ff2d6322d13448db8cfd5fb9267517ab4b96d0d299ed45fR137).

node/src/command.rs
```diff
@@ -451,7 +460,7 @@ pub fn run() -> Result<()> {
 						sp_core::crypto::set_default_ss58_version(
 							amplitude_runtime::SS58Prefix::get().into(),
 						);
-						crate::service::start_parachain_node_spacewalk_amplitude::<AmplitudeRuntimeExecutor>(
```
TorstenStueber, line 454: I could remove the generic type parameters from the functions `start_parachain_node_spacewalk_amplitude`, etc.

node/src/rpc.rs
```diff
@@ -44,26 +38,12 @@ pub struct FullDeps<C, P> {
 }
 
 /// Instantiate all RPC extensions.
-pub fn create_full_pendulum<C, P>(
-	deps: FullDeps<C, P>,
+pub fn create_full_pendulum<P>(
```
TorstenStueber, line 41: All the functions in this file were unnecessarily generic. This is not required – they are used for exactly one of the four runtimes. Thus, also the complex type guards could be removed.

node/src/service.rs
```diff
@@ -37,11 +37,50 @@ use substrate_prometheus_endpoint::Registry;
...
+	create_full_amplitude, create_full_development, create_full_foucoco, create_full_pendulum,
+	FullDeps, RpcExtension,
+};
+
+pub use amplitude_runtime::RuntimeApi as AmplitudeRuntimeApi;
+pub use development_runtime::RuntimeApi as DevelopmentRuntimeApi;
+pub use foucoco_runtime::RuntimeApi as FoucocoRuntimeApi;
+pub use pendulum_runtime::RuntimeApi as PendulumRuntimeApi;
+
+pub type AmplitudeClient =
```
TorstenStueber, line 50: It helps a lot to define these types because they are used frequently throughout the code.

node/src/service.rs
```diff
@@ -37,11 +37,50 @@ use substrate_prometheus_endpoint::Registry;
...
+pub type DevelopmentClient =
+	TFullClient<Block, DevelopmentRuntimeApi, NativeElseWasmExecutor<DevelopmentRuntimeExecutor>>;
+
 type ParachainBlockImport<RuntimeApi, Executor> = TParachainBlockImport<
 	Block,
 	Arc<TFullClient<Block, RuntimeApi, NativeElseWasmExecutor<Executor>>>,
 	TFullBackend<Block>,
 >;
+
+pub trait ParachainRuntimeApiImpl:
```
TorstenStueber, line 65: This is a type guard that was used a lot in the functions in this file. Since this type guard is satisfied by all runtime api implementations of each of the four runtimes, we could simplify the type guards by just defining a new trait and then express that the four runtimes implement this trait.

node/src/service.rs
```diff
@@ -238,23 +265,34 @@ async fn build_relay_chain_interface(
 	}
 }
 
-// We use this as the RuntimeApi definition in the following function so that we don't need to specify so many generic traits.
-// This might have to change if we want to use this function for other runtimes.
-use amplitude_runtime::RuntimeApi as AmplitudeRuntimeApi;
+type FullDepsOf<RuntimeApi, Executor> = FullDeps<
```
TorstenStueber, line 268: Just a type that is used in the function `start_node_impl`, but is so complex that I wanted to define here.

node/src/service.rs
```diff
@@ -238,23 +265,34 @@ async fn build_relay_chain_interface(
...
+	sc_transaction_pool::FullPool<
+		Block,
+		TFullClient<Block, RuntimeApi, NativeElseWasmExecutor<Executor>>,
+	>,
+>;
 
-/// Start a node that exposes the RPC endpoints for the spacewalk pallets for the amplitude runtime.
 #[sc_tracing::logging::prefix_logs_with("Parachain")]
-async fn start_node_impl_amplitude<Executor>(
+async fn start_node_impl<RuntimeApi, Executor>(
```
TorstenStueber, line 277: All the four `start_node_impl_<...>` functions were *exactly* the same with the exception being a single line that calls either one of the four functions in `rpc.rs`. I now added the respective function as an extra argument and define which of the functions to use when calling `start_node_impl` in any one of the four `start_parachain_node_<...>` functions at the end of this file.

node/src/service.rs
```diff
@@ -313,13 +351,10 @@ where
...
-			let deps = crate::rpc::FullDeps {
-				client: client.clone(),
-				pool: transaction_pool.clone(),
-				deny_unsafe,
-			};
+			let deps =
+				FullDeps { client: client.clone(), pool: transaction_pool.clone(), deny_unsafe };
 
-			crate::rpc::create_full_amplitude(deps).map_err(Into::into)
+			create_full_rpc(deps).map_err(Into::into)
```
TorstenStueber, line 357: This is the occurrence of the one function that has been different between the four `start_node_impl_<...>` functions.

node/src/service.rs
```diff
@@ -414,669 +449,86 @@ where
...
-	polkadot_config: Configuration,
-	collator_options: CollatorOptions,
-	id: ParaId,
-	hwbench: Option<sc_sysinfo::HwBench>,
-) -> sc_service::error::Result<(
-	TaskManager,
-	Arc<TFullClient<Block, FoucocoRuntimeApi, NativeElseWasmExecutor<Executor>>>,
-)>
+/// Build the import queue for the parachain runtime.
+fn build_import_queue<RuntimeApi, Executor>(
```
TorstenStueber, line 453: The diff algorithm gets a little confused here and does not properly align the old and new code. `build_import_queue` occurs below the four `start_node_impl_<...>` functions and is basically unchanged. I could just use the greatly simplified type guard `RuntimeApi::RuntimeApi: ParachainRuntimeApiImpl`.

node/src/service.rs
```diff
@@ -1133,112 +585,77 @@ where
 }
 
 /// Start a parachain node.
-pub async fn start_parachain_node_pendulum<RuntimeApi, Executor>(
+pub async fn start_parachain_node_pendulum(
```
TorstenStueber, line 588: These last four functions are now much cleaner because they were unnecessarily generic.

ebma:
Is this ready for review or still WIP? Just asking because no reviewers were added yet.

TorstenStueber:
This is ready for review.

Is the workflow that we add reviewers manually?

ebma:
> Is the workflow that we add reviewers manually?

Yes, once ready you'd pick some reviewers so that it's clear that you want to have a review. Otherwise, people would miss it since they don't get any notifications if they don't 'watch' the repository.

ebma (approved):
Really amazing changes, this simplifies a lot :+1:

b-yap (approved):

ebma (approved):

Issue #286, opened 2023-07-26
Implement new RPC function of Spacewalk's 'oracle' pallet for all runtimes


prayagd:
## Requirement
* This is a follow ticket for this [one](https://github.com/pendulum-chain/spacewalk/issues/383)
* Configure new RPC function in all runtimes (Foucoco, Amplitude and Pendulum) in the following order
* Runtime upgrade needed on all three networks
* Re-deploy the nodes of all three networks for changes to take effect.

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/pendulum/286?planning-poker) with Zenhub @adelarja @ashneverdawn @b-yap @ebma @TorstenStueber

PR #311, opened 2023-10-03
Implement new rpc function of spacewalks oracle pallet for all runtimes

gianfra-t:
Note: Requires merge spacewalk's [PR #402](https://github.com/pendulum-chain/spacewalk/pull/402), For now I will leave as draft to later point to the corresponding commit rev of spacewalk.

Implements the new RPC function to get exchange rates for Foucoco, Amplitude.
Pendulum runtime does not currently implements `oracle` pallet, this would be required first to implement the new RPC method.

Issue #286

This will require a node redeployment since runtime is modified.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -191,6 +191,7 @@ version = "0.1.0"
 dependencies = [
  "bifrost-farming",
  "bifrost-farming-rpc-runtime-api",
+ "clients-info",
  "cumulus-pallet-aura-ext",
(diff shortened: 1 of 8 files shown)


ebma (approved):
Looks good 👍 
I'll merge this now because the PR that originally blocked it was already merged and I added the same changes in [this](https://github.com/pendulum-chain/pendulum/pull/323) PR already (because I forgot that this one existed).

PR #332, opened 2023-10-27
315 enable pendulum assets to be transferable via xcm

bogdanS98:
Changes in this PR:

Enable native token XCM transfers for all runtimes:

- [x] Pendulum 
- [x] Amplitude
- [x] Foucoco
- [x] Integration tests


runtime/integration-tests/src/test_macros.rs
```diff
@@ -461,9 +461,97 @@ macro_rules! parachain1_transfer_asset_to_parachain2_and_back {
...
+		// Verify the balance on the AssetHub chain after transfer
+		// $assethub_chain::execute_with(|| {
+		// 	   // I need ForeignAssets here instead of Assets
+		//     use $assethub_runtime::Assets;
+
+		//     assert_eq!(
+		//         Assets::balance(asset_location, &BOB.into()),
+		//         transfer_amount
+		//     );
+		// });
```
ebma, line 539: Using 
```
use $assethub_runtime::ForeignAssets;
ForeignAssets::balance(asset_location, &BOB.into()),
```
doesn't work?

bogdanS98: No, unfortunately

ebma: Okay nevermind, of course it does not work because the asset_hub dependency we are using is [this](https://github.com/pendulum-chain/pendulum/blob/9d5791f2b12db461b3d7594f2a9e61d7d534ac10/runtime/integration-tests/Cargo.toml#L39C30-L39C67) one.

And ForeignAssets does not exist [here](https://github.com/paritytech/cumulus/blob/e05c8d7f71734ed71188337c6cb0d30715f6320f/parachains/runtimes/assets/statemint/src/lib.rs#L635).

We would probably need to replace the `statemint` and `statemine` dependencies we are defining in the Cargo.toml with the more recent asset hubs [here](https://github.com/paritytech/polkadot-sdk/tree/master/cumulus/parachains/runtimes/assets). Can you try that @bogdanS98? I am rather pessimistic that this works though as this will probably cause compilation issues with cargo due to dependency mismatches. Still worth a shot though.

runtime/pendulum/src/xcm_config.rs
```diff
@@ -95,6 +99,9 @@ impl Convert<MultiLocation, Option<CurrencyId>> for CurrencyIdConvert {
 						GeneralIndex(asset_hub::USDT_ASSET_ID),
 					),
 			} => Some(CurrencyId::XCM(XCM_ASSET_ASSETHUB_USDT)),
+			MultiLocation { parents: 1, interior: X2(Parachain(id), PalletInstance(10)) }
```
TorstenStueber, line 102: Are we confident that this is correct (is it defined this way on other chains or do the XCM docs specify that we need to do it this way?).

MultiLocations are always relative and I know that they are usually transformed when they occur in an XCM messages that is sent to another chain so that the MultiLocation that made sense in the context of the sender of the parachain will then look different and will be adapted to the context of the target parachain where it is executed.

In this case I think that an incoming XCM messages that is sent to us will actually have a multilocation that will be transformed to look like `MultiLocation { parents: 0, interior: X1(PalletInstance(10)) }`.

I could be totally wrong, though.

bogdanS98: You're right, I also added `MultiLocation { parents: 0, interior: X1(PalletInstance(10)) }` because it is needed when receiving the native token via XCM. Sender uses `MultiLocation { parents: 1, interior: X2(Parachain(id), PalletInstance(10)) }` but this gets transformed on the receiving side before being converted to CurrencyID.

runtime/integration-tests/src/pendulum_tests.rs
```diff
@@ -116,3 +129,16 @@ fn assethub_transfer_asset_to_pendulum_and_back() {
 		network_id
 	);
 }
+
+#[test]
+fn transfer_native_token_to_sibling_parachain_and_back() {
```
ebma, line 134: Nit: let's use explicit parachain names here so that it's easier to distinguish the tests of pendulum and amplitude. I would suggest we use `transfer_native_token_from_pendulum_to_sibling_parachain_and_back()`

runtime/integration-tests/src/sibling.rs
```diff
@@ -0,0 +1,557 @@
...
+
+impl Convert<MultiLocation, Option<CurrencyId>> for CurrencyIdConvert {
+	fn convert(location: MultiLocation) -> Option<CurrencyId> {
+		match location {
+			// Just for testing purposes, parachain id is not verified so this can be used by all runtimes
+			MultiLocation { parents: 1, interior: X2(Parachain(_id), PalletInstance(10)) } =>
+				Some(CurrencyId::Native),
+			MultiLocation { parents: 0, interior: X1(PalletInstance(10)) } =>
+				Some(CurrencyId::Native),
+			_ => None,
```
ebma, line 93: This means that any native tokens that we receive from another chain are mixed up with the native tokens we already have on the sibling chain, no?  I would prefer we make this conversion very explicit and use a new `CurrencyId` that is custom to the sibling chain.

I'd create a `CurrencyId` similar to this
```
pub enum CurrencyId {
 Relay,
 Pendulum,
 Amplitude
}
```

and then in the conversion functions we do something like:
```
impl Convert<MultiLocation, Option<CurrencyId>> for CurrencyIdConvert {
	fn convert(location: MultiLocation) -> Option<CurrencyId> {
              match location {
			MultiLocation { parents: 1, interior: X2(Parachain(PENDULUM_ID), PalletInstance(10)) } =>
				Some(CurrencyId::Pendulum),
                         MultiLocation { parents: 1, interior: X2(Parachain(AMPLITUDE_ID), PalletInstance(10)) } =>
				Some(CurrencyId::Amplitude),
			MultiLocation { parents: 0, interior: X1(PalletInstance(10)) } =>
				Some(CurrencyId::Native),
			_ => None,
```

and

```
impl Convert<CurrencyId, Option<MultiLocation>> for CurrencyIdConvert {
	fn convert(id: CurrencyId) -> Option<MultiLocation> {
		match id {
			CurrencyId::Native => Some(MultiLocation::new(
				1,
				X2(Parachain(ParachainInfo::parachain_id().into()), PalletInstance(10)),
			)),
			CurrencyId::Pendulum => Some(MultiLocation::new(
				1,
				X2(Parachain(PENDULUM_ID).into()), PalletInstance(10)),
			)),
			CurrencyId::Amplitude => Some(MultiLocation::new(
				1,
				X2(Parachain(AMPLITUDE_ID).into()), PalletInstance(10)),
			)),
			_ => None,
		}
	}
}
```

ebma (changes requested):

runtime/integration-tests/src/sibling.rs
```diff
@@ -0,0 +1,613 @@
...
+}
+
+// Convert from u32 parachain id to CurrencyId
+// Needed for the test macro so it works regardless of the XCM sender parachain
+impl From<u32> for CurrencyId {
+	fn from(id: u32) -> Self {
+		match id {
+			PENDULUM_ID => CurrencyId::Pendulum,
+			AMPLITUDE_ID => CurrencyId::Amplitude,
+			KUSAMA_ASSETHUB_ID | POLKADOT_ASSETHUB_ID => CurrencyId::XCM(XCM_ASSET_ASSETHUB_USDT),
```
bogdanS98, line 103: ATM both Kusama and Polkadot asset hub have the same ids and this gives a warning of "unreachable" for the second pattern. I left it here in case we change the ids at some point. In that case, the warning will disappear.

runtime/integration-tests/src/sibling.rs
```diff
@@ -0,0 +1,613 @@
...
+				parents: 1,
+				interior: X2(Parachain(PENDULUM_ID), PalletInstance(10)),
+			} => Some(CurrencyId::Pendulum),
+			MultiLocation {
+				parents: 1,
+				interior: X2(Parachain(AMPLITUDE_ID), PalletInstance(10)),
+			} => Some(CurrencyId::Amplitude),
+			MultiLocation { parents: 0, interior: X1(PalletInstance(10)) } =>
+				Some(CurrencyId::Native),
+			// TODO see how to use KUSAMA/POLKADOT asset hub based on the XCM sender chain (PENDULUM/AMPLITUDE) if needed
```
bogdanS98, line 148: Here too

ebma (approved):
Looks good to me now 👍

PR #335, opened 2023-11-06
add new reward/reward-distribution pallets to foucoco and amplitude

### LINKED ISSUE #314: Add the new Spacewalk pallets
This ticket is a part of https://github.com/pendulum-chain/tasks/issues/47

Spacewalk adds two new pallets:
* pooled-rewards
* reward-distribution

These pallets should be added to the Foucoco and Amplitude implementation of Spacewalk and should use the following configuration parameters:

## Pooled rewards
* set `PoolId` configuration to `CurrencyId`
* set `StakeId` configuration to `AccountId`

## Reward-distribution
* set `Currency` to pallet-balances

gianfra-t:
Closes #314. 

Adds the new `pooled-reward` and `reward-distribution` pallets to Foucoco and Amplitude runtimes.

As was discussed previously in [the implementation](https://github.com/pendulum-chain/spacewalk/pull/415), `PoolId` is set to `CurrencyId` and `StakeId` is set to `VaultId`

For now the new value of `MaxRewardCurrencies` of the staking pallet is set to 10.

### DIFF
--- a/runtime/amplitude/src/lib.rs
+++ b/runtime/amplitude/src/lib.rs
@@ -358,7 +358,7 @@ impl Contains<RuntimeCall> for BaseFilter {
 			RuntimeCall::Security(_) |
 			RuntimeCall::StellarRelay(_) |
 			RuntimeCall::VaultRegistry(_) |
-			RuntimeCall::VaultRewards(_) |
+			RuntimeCall::PooledVaultRewards(_) |
 			RuntimeCall::Farming(_) |
 			RuntimeCall::AssetRegistry(_) |
 			RuntimeCall::RewardDistribution(_)=> true,
@@ -1152,7 +1152,7 @@ impl fee::Config for Runtime {
 	type SignedInner = SignedInner;
 	type UnsignedFixedPoint = UnsignedFixedPoint;
 	type UnsignedInner = UnsignedInner;
-	type VaultRewards = VaultRewards;
+	type VaultRewards = PooledVaultRewards;
 	type VaultStaking = VaultStaking;
 	type OnSweep = currency::SweepFunds<Runtime, FeeAccount>;
 	type MaxExpectedValue = MaxExpectedValue;
@@ -1202,7 +1202,9 @@ impl clients_info::Config for Runtime {
 	type MaxNameLength = ConstU32<255>;
 	type MaxUriLength = ConstU32<255>;
 }
-
+// Choice of parameters: Perquintill::from_parts(37567400000000000u64) represents a value of
+// 0.0375674 = 37567400000000000 / 1×10¹⁸
+// The decay interval 216000 equates to a month when considering 1 block every 12 seconds
 parameter_types! {
 	pub const DecayRate: Perquintill = Perquintill::from_parts(37567400000000000u64);
 	pub const MaxCurrencies: u32 = 10;
@@ -1214,7 +1216,7 @@ impl reward_distribution::Config for Runtime {
 	type Balance = Balance;
 	type DecayInterval = ConstU32<216_000>;
 	type DecayRate = DecayRate;
-	type VaultRewards = VaultRewards;
+	type VaultRewards = PooledVaultRewards;
 	type MaxCurrencies = MaxCurrencies;
 	type OracleApi = Oracle;
 	type Balances = Balances;
@@ -1318,7 +1320,7 @@ construct_runtime!(
 		Security: security::{Pallet, Call, Config, Storage, Event<T>} = 67,
 		StellarRelay: stellar_relay::{Pallet, Call, Config<T>, Storage, Event<T>} = 68,
 		VaultRegistry: vault_registry::{Pallet, Call, Config<T>, Storage, Event<T>, ValidateUnsigned} = 69,
-		VaultRewards: pooled_rewards::{Pallet, Call, Storage, Event<T>} = 70,
+		PooledVaultRewards: pooled_rewards::{Pallet, Call, Storage, Event<T>} = 70,
 		VaultStaking: staking::{Pallet, Storage, Event<T>} = 71,
 		ClientsInfo: clients_info::{Pallet, Call, Storage, Event<T>} = 72,
 		RewardDistribution: reward_distribution::{Pallet, Call, Storage, Event<T>} = 73,
--- a/runtime/amplitude/src/lib.rs
+++ b/runtime/amplitude/src/lib.rs
@@ -1204,7 +1204,7 @@ impl clients_info::Config for Runtime {
 }
 
 parameter_types! {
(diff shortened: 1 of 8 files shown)


runtime/amplitude/src/lib.rs
```diff
@@ -1201,6 +1203,35 @@ impl clients_info::Config for Runtime {
 	type MaxUriLength = ConstU32<255>;
 }
 
+parameter_types! {
+	pub const DecayRate: Perquintill = Perquintill::from_percent(5);
```
ebma, line 1207: We need to change this value to `0.0375674` ie 3.75674 percent. So we probably need to do something like `Perquintill::from_rational(375_674, 10_000_000)` (?)
according to https://github.com/pendulum-chain/tasks/issues/47

runtime/amplitude/src/lib.rs
```diff
@@ -1201,6 +1203,35 @@ impl clients_info::Config for Runtime {
...
+	pub const MaxCurrencies: u32 = 10;
+}
+
+impl reward_distribution::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type WeightInfo = reward_distribution::SubstrateWeight<Runtime>;
+	type Balance = Balance;
+	type DecayInterval = ConstU32<100>;
```
ebma, line 1215: The decay interval should be set to 1 month, according to https://github.com/pendulum-chain/tasks/issues/47
Since we can only measure it in blocks, we have to assume that a decay interval of `216_000` blocks roughly relates to a month, assuming a month consists of 30 days.

runtime/amplitude/src/lib.rs
```diff
@@ -1287,9 +1318,10 @@ construct_runtime!(
 		Security: security::{Pallet, Call, Config, Storage, Event<T>} = 67,
 		StellarRelay: stellar_relay::{Pallet, Call, Config<T>, Storage, Event<T>} = 68,
 		VaultRegistry: vault_registry::{Pallet, Call, Config<T>, Storage, Event<T>, ValidateUnsigned} = 69,
-		VaultRewards: reward::{Pallet, Call, Storage, Event<T>} = 70,
+		VaultRewards: pooled_rewards::{Pallet, Call, Storage, Event<T>} = 70,
```
ebma, line 1321: I'm actually not 100% sure if it leads to issues if we replace this pallet with another one. Let's be on the safe side and not assign the index of `70` here but use `74`.

ebma (changes requested):
Looks good overall but we should use other values. The comments apply to both amplitude and foucoco.

runtime/amplitude/src/lib.rs
gianfra-t: I totally forgot to change the values according to the task. Yes it should be that, I will double check with the docs.

runtime/amplitude/src/lib.rs
gianfra-t: But from the perspective of whatever was already using the old rewards pallet, it would look like it disappeared. Are you sure the safe side this time means changing it to 74?

runtime/amplitude/src/lib.rs
gianfra-t: I actually had to go with `from_parts` since apparently `from_rational` is not allowed in this scope. 
```parameter_types! {
	pub const DecayRate: Perquintill = Perquintill::from_parts(37567400000000000u64);
```
If we divide that number by the maximum of the type (which I assume means the quintillion) we get `37567400000000000 / 1000000000000000000 = 0.0375674`

ebma: Nice, looks good

ebma: We could also think of adding comments above the new definitions that say that the value means `0.0375674` and 1 month respectively. Maybe not obvious at first.

runtime/amplitude/src/lib.rs
ebma: No, I'm not. I know that the order matters to some extend so I tried to look it up. In the last paragraph of [this](https://docs.substrate.io/reference/frame-macros/#runtime-construction-macros) section it is mentioned that it is important that a pallet that depends on another pallet comes after the pallet it depends on. So according to this, it would indeed be a mistake to assign `74` to the `VaultRewards` pallet since `RewardDistribution` depends on it.

What I'm still not sure about is whether we can just re-use indices. 

> But from the perspective of whatever was already using the old rewards pallet, it would look like it disappeared

I'm more concerned about the storage items of the pallet disappearing or rather _not disappearing_ and being re-used by our new pooled rewards pallet. I fear that the only way to test if the configuration you chose causes any issues is by testing the runtime upgrade locally

gianfra-t: Thanks for the explanation! I still don't quite understand why the storage itself would be a problem beacuse I am looking at the docs but it says right before the dependency part:

> the order used for defining pallets in the construct_runtime! macro affects the genesis storage configuration. 

In the [rust docs](https://paritytech.github.io/polkadot-sdk/master/frame_support/macro.construct_runtime.html) it is also stated that:
> The population of the genesis storage depends on the order of pallets. 

This makes me think a runtime upgrade wouldn't be a problem I guess? Since we are not going through the genesis config and also these pallets do not have one.

ebma: So what I am worried about is that our new pooled rewards pallet accesses storage items of the older rewards pallet that we used previously. If you look [here](https://polkadot-blockchain-academy.github.io/pba-book/frame/storage/page.html#frame-storage-keys), the pallet name is used to derive the key to the storage item in the database. And if our new pooled rewards pallet shares the same name as the old one (ie. `VaultRewards`), and it also uses the same names for some of its storage items (which it does for almost all storage items, see [here](https://github.com/pendulum-chain/spacewalk/blob/main/pallets/reward/src/lib.rs#L101-L127) vs [here](https://github.com/pendulum-chain/spacewalk/blob/main/pallets/pooled-rewards/src/lib.rs#L119-L162)), then our new pallet will query data of the old pallet which it should not. 

However it should be fine if we just rename the pallet instance in our runtime. So instead of `VaultRewards` we can call it `PooledVaultRewards` or something.

ebma: About the indices, I think you are right and we don't need to worry about it. @b-yap found [this](https://substrate.stackexchange.com/questions/10428/reusing-old-pallet-instance-number/10429#10429) related forum post which indicates that this is not problematic as long as we don't have any pending transactions scheduled.

gianfra-t: So let's rename it to be on the safe side

ebma (approved):
Looks good to me now 👍

ebma (approved):

PR #342, opened 2023-11-09
New currency type

### LINKED ISSUE #334: Use latest Spacewalk CurrencyId in runtimes
Following the creation of the new asset type, CurrencyID, we need to incorporate it into the Pendulum runtimes

## Requirements:

Update the Pendulum runtimes to include the newly defined CurrencyID asset type.

gianfra-t:
Closes #334. 

The new type `CurrencyId::Token(u64)` was added to spacewalk [here](https://github.com/pendulum-chain/spacewalk/pull/442).

The new currency type is disabled to zenlink conversion functions.


ebma (commented):
Thanks for the nice description of the PR @gianfra-t. 
It seems like we only have a choice between two 'easy' solutions:
1) we disallow using `CurrencyId::Token`s in Zenlink pools completely or 
2) we limit it to the `CurrencyId::Token`s that fit into one byte, like you implemented it. 

Everything else is probably more complicated because we need to change the ZenlinkLPToken type. 

It might be worth discussing these two approaches before we choose one over the other. For now, I would actually vote to go for approach 1) because this way we still have time to think about the 'ideal' approach until we actually need it. If we use approach 2) and a pool is created with this we might limit ourselves in the future if we choose to come up with some fancy custom derivation function of the ID or whatever.

What's your opinion on this @gianfra-t? Maybe also @b-yap, @bogdanS98  and @TorstenStueber

gianfra-t:
If we are not interested in using Zenlink with these new currency type now, then I also think it is better to just not allow it and in the near future think of another way to derive the IDs not just for this new type but for other types that we could add, according to our future needs. 
Because even if we change the internal type of `CurrencyId::ZenlinkLPToken(u8, u8, u8, u8)` to `u256`, we will still be limited by the 2 byte representation of the `asset_index` shared through all currency types.

ebma:
True. Alright then let's completely disallow using `Token` for Zenlink pools and postpone thinking of a derivation once needed.

runtime/common/src/zenlink.rs
```diff
@@ -235,6 +236,8 @@ mod zenlink_tests {
 			zenlink_id_to_currency_id(fake_zenlink_asset, 1000u32);
 		let expected_currency: CurrencyId = CurrencyId::ZenlinkLPToken(0, 2, 1, 2);
 		assert_eq!(currency, Ok(expected_currency));
+
+
```
ebma, line 240: Nit: unnecessary new lines

runtime/common/src/zenlink.rs
```diff
@@ -98,6 +98,7 @@ pub fn currency_id_to_zenlink_id(
 				((token2_type as u64) << 40);
 			Ok(ZenlinkAssetId { chain_id: parachain_id, asset_type: LOCAL, asset_index: index })
 		},
+		_ => Err(()),
```
ebma, line 101: Nit: We could also make it explicit here by creating an extra match branch for the `Token()` and returning an error. This way, we kind of show that we deliberately decided not to support it.

ebma (approved):
LGTM overall

runtime/common/src/zenlink.rs
gianfra-t: I see, also maintaining the `_=>` branch for extra cases. Seems good.

ebma (approved):

Issue #249, opened 2023-06-21
Extract `chain_extension` implementation and re-use it in all runtimes


ebma:
We want to use the same `chain_extension` implementation in all our runtimes. In order not to duplicate any code, we should extract the implementation of the `chain_extension` into another directory, similar to what Astar did [here](https://github.com/AstarNetwork/Astar/tree/master/chain-extensions), and re-use that in each runtime.

To add a bit more structure, we can even split our current chain extension into two sub-parts: a 'token-extension' which is comprised of the functions with ID 1100-1199 and a 'price-extension' which consists of functions with ID 1200-1299. Astar also found a way to use register three chain extensions at the same time, see [here](https://github.com/AstarNetwork/Astar/blob/51873938c8d63cc87885db95af74cfb3ec36d015/runtime/shibuya/src/lib.rs#L683-L687).

ebma:
@pendulum-chain/product this ticket should be prioritized soon. Having this for before Nabla launch would be good, because otherwise we have to duplicate code once we want to use our Solidity wrapper contract solution also on Pendulum.

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/pendulum/249?planning-poker) with Zenhub @adelarja @b-yap @ebma @TorstenStueber

TorstenStueber:
@prayagd @ebma can we move this ticket to Ready? Any reason why not?

ebma:
We can move it already but should only work on it once #356 is merged.

prayagd:
@ebma just to check this is important for the Nabla mainnet launch right?

ebma:
It's a chore and so is not a stopper for anything. It's about refactoring some code and wouldn't add or remove any functionalities. We should still do it eventually and I wouldn't like to have it in the icebox.

PR #348, opened 2023-11-14
260 audit pdm 008 hardcoded constants in match

### LINKED ISSUE #260: Audit: PDM-008 - Hardcoded Constants in match
The presence of hardcoded constants in the match statement within the implementation of ChainExtension hampers code readability.

ID PDM-008
Scope Code Quality
Status New

**Description**
The following code segment has drawn our attention: `runtime/foucoco/src/lib.rs:964`:
```
match func_id { //transfer
     1105 => { /* ... */ },
     //balance
     1106 => { /* ... */ },
     //total_supply
     1107 => { /* ... */ },
     //approve_transfer
     1108 => { /* ... */ },
     //transfer_approved
     1109 => { /* ... */ },
     //allowance
     1110 => { /* ... */ },
     //dia price feed
     7777 => { /* ... */ },
     _ => { /* ... */ },
}
```
While the code comments provide some understanding of the functionality associated with each func_id , it is recommended to enhance
the code logic in this section to improve comprehensibility, rather than solely relying on comments.

**Recommendation**
We recommend creating a separate enum to handle all the supported func_id options, using descriptive names that correspond to each function. This approach enhances code readability and maintainability:
```
enum FuncId { Transfer,
     Balance,
     TotalSupply
     ApproveTransfer,
     TransferApproved,
     Allowance,
     DiaPriceFeed,
}
impl TryFrom<u16> for FuncId {
     type Error = DispatchError;
     fn try_from(func_id: u16) -> Result<Self, Self::Error> {
          let id = match func_id {
               1105 => Self::Transfer,
               1106 => Self::Balance,
               1107 => Self::TotalSupply, 
               1108 => Self::ApproveTransfer, 
               1109 => Self::TransferApproved, 
               1110 => Self::Allowance,
               7777 => Self::DiaPriceFeed,
                _ => {
                      error!("Called an unregistered `func_id`: {:}", func_id);
                      return Err(DispatchError::Other("Unimplemented func_id")) 
               }
          };
          Ok(id) 
     }
}
```
This enum allows for easier handling of the logic within the match statement in the call implementation.
Furthermore, it is good practice to implement the logic for each function in a separate method instead of encapsulating all the code within the match arms. This promotes code modularity and readability.
Here's an example of how the code would look with the recommended changes:
```
let func_id = FuncId::try_from(env.func_id())?; match func_id {
     FuncId::Transfer => /* Call method that performs transfer */,
     FuncId::Balance => /* Call method that returns balance */,
     FuncId::TotalSupply => /* ... */ ,
     FuncId::ApproveTransfer => /* ... */,
     FuncId::TransferApproved => /* ... */,
     FuncId::Allowance => /* ... */,
     FuncId::DiaPriceFeed => /* ... */,
 }
```
With these improvements, the code becomes more comprehensible, maintainable, and follows best practices.

bogdanS98:
Closes #260

### DIFF
--- a/runtime/foucoco/src/lib.rs
+++ b/runtime/foucoco/src/lib.rs
@@ -374,7 +374,7 @@ impl Contains<RuntimeCall> for BaseFilter {
 			RuntimeCall::TokenAllowance(_) |
 			RuntimeCall::AssetRegistry(_) |
 			RuntimeCall::Proxy(_) |
-			RuntimeCall::RewardDistribution(_)=> true,
+			RuntimeCall::RewardDistribution(_) => true,
 			// All pallets are allowed, but exhaustive match is defensive
 			// in the case of adding new pallets.
 		}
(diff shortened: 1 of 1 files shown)


runtime/foucoco/src/lib.rs
```diff
@@ -973,14 +1011,14 @@ where
...
+	fn call<E: Ext>(&mut self, env: Environment<E, InitState>) -> Result<RetVal, DispatchError>
 	where
 		E: Ext<T = T>,
 		<E::T as SysConfig>::AccountId: UncheckedFrom<<E::T as SysConfig>::Hash> + AsRef<[u8]>,
 	{
-		let func_id = env.func_id();
+		let func_id = FuncId::try_from(env.func_id())?;
 
-		trace!("Calling function with ID {} from Psp22Extension", func_id);
+		trace!("Calling function with ID {} from Psp22Extension", env.func_id());
```
ebma, line 1021: Nit: Can we just clone `func_id` that we assigned previously?

ebma (commented):
The code looks good overall. However, now that we extract the function calls into separate functions, we could also put them into an extra file to clean up the structure. This would be similar to the task defined in [this](https://github.com/pendulum-chain/pendulum/issues/249) ticket though. For #249 we'd have to do more than putting this into an extra file though, so I'm not sure if it makes sense to do both in this PR.

Maybe it's best if we leave all functions here as you did. But let's at least order them by the way they are called and defined in the `FuncId` enum.

ebma (approved):
LGTM. Can be merged once CI passed

PR #349, opened 2023-11-16
250 improve error handling in chain extension

### LINKED ISSUE #250: Improve error handling in chain extension
At the moment, we throw a lot of `DispatchError::Other()` errors in our chain extension. 
When a smart contract calls our chain extension and such an error is shown, the whole execution traps and no error message is shown (because although `DispatchError::Other()` contains a message, this messages is stripped).
What we should do instead is, to create an error enum (like [this](https://github.com/AstarNetwork/Astar/blob/master/chain-extensions/types/dapps-staking/src/lib.rs)) and in case we encounter an error in our chain extension function, instead of aborting with a `DispatchError`, we encode the respective error (like [this](https://github.com/AstarNetwork/Astar/blob/51873938c8d63cc87885db95af74cfb3ec36d015/chain-extensions/dapps-staking/src/lib.rs#L249)) and send it back to our smart contract, which is then able to show a proper error message based on that index.

This would improve the developer experience, because, at the moment you don't see what exactly failed about your contract call.

gianfra-t:
Closes #250 . 

This PR proposes to use the existing `ChainExtensionError` as the [enum](https://github.com/AstarNetwork/Astar/blob/master/chain-extensions/types/dapps-staking/src/lib.rs#L26) discussed in the issue, which is now renamed `ChainExtensionOutcome` which also holds the `Success` variant.

For this, each value is [mapped](https://github.com/pendulum-chain/pendulum/pull/349/files#diff-e9470b62a3b87e17d0c21ab8c16a0a7afca151bbfda1cfe75bfe3f3126a8b274R185) to a corresponding integer, and later we [use](https://github.com/pendulum-chain/pendulum/pull/349/files#diff-63fad4ac7944e1d379fd236f19986bfb4f8c39eb199a14535775de9a81d72a08R1095) the corresponding `ChainExtensionOutcome` as `u32` when returning.

A conversion from `u32` to `ChainExtensionOutcome` or it's inner variants is also [defined](https://github.com/pendulum-chain/pendulum/pull/349/files#diff-e9470b62a3b87e17d0c21ab8c16a0a7afca151bbfda1cfe75bfe3f3126a8b274R232)


gianfra-t:
I understand the conversion back from number to chain extension error, and I think also returning a success is more readable than 0. Then I would propose to rename the enum to something that is not `ChainExtensionError`, so that it does not become confusing (naming it `Outcome`, let's say)

runtime/common/src/chain_ext.rs
```diff
@@ -174,3 +180,49 @@ pub fn decode<T: Decode>(input: Vec<u8>) -> Result<T, codec::Error> {
 	let mut input = input.as_slice();
 	T::decode(&mut input)
 }
+
+impl ChainExtensionError {
+    pub fn as_u32(&self) -> u32 {
+        match self {
+            ChainExtensionError::Other => 1,
```
b-yap, line 187: just curious: any reason why the number doesn't start at 0?

ebma: Yeah, the index of this error will be returned to the calling contract with [RetVal::Converging(index)](https://crates.parity.io/pallet_contracts/chain_extension/enum.RetVal.html). And a `0` would generally indicate a success. That's why it makes sense that the errors start at `1`.

ebma:
> Then I would propose to rename the enum to something that is not ChainExtensionError, so that it does not become confusing (naming it Outcome, let's say)

Sure. I would be fine with `ChainExtensionOutcome`. We could also do `ChainExtensionRetVal` or `ChainExtensionReturnValue`. Not sure what's best, WDYT @gianfra-t and @b-yap?

runtime/common/src/chain_ext.rs
gianfra-t: Yeah we should definitely call it ChainExtensionOutcome and start at 0 with success. Otherwise it's too confusing.

gianfra-t:
I've added the conversion from u32 to `ChainExtensionOutcome` and adjusted the code to the refactor made in a recently merged PR.

ebma:
I think the CI basically passed because only saving the cache failed, right? Maybe we can just merge it like this

gianfra-t:
Does that step in the CI just cleans after the cargo build? In that case yes we should already merge it. Otherwise we can re-check again.

ebma:
I think it's fine to ignore. I'll merge it now.

PR #350, opened 2023-11-17
Expose parachain-staking methods to RPC in all runtimes

gianfra-t:
Please ensure that the collator nodes are redeployed after this PR is merged.  

This PR adds  `get_unclaimed_staking_rewards` and `get_staking_rates` methods from `parachain-staking` to the RPC configuration and allows them to be called.


pallets/parachain-staking/rpc/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+authors = ["Interlay <contact@interlay.io>"]
```
ebma, line 2: Interlay is not the author of this. It should rather refer to KILT

pallets/parachain-staking/rpc/runtime-api/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+authors = ["Interlay <contact@interlay.io>"]
```
ebma, line 2: Same here

res/amplitude-spec-raw.json
ebma, line 1: Let's not touch any of the spec files. They are important for the genesis and if we change them now and want to have new collator nodes join the network, they wouldn't be able to sync as they would generate a different genesis block than is in our archives.

runtime/amplitude/src/lib.rs
```diff
@@ -1524,12 +1524,13 @@ impl_runtime_apis! {
 		}
 	}
 
-	impl parachain_staking::runtime_api::ParachainStakingApi<Block, AccountId, Balance> for Runtime {
-		fn get_unclaimed_staking_rewards(account: &AccountId) -> Balance {
-			ParachainStaking::get_unclaimed_staking_rewards(account)
+	impl module_pallet_staking_rpc_runtime_api::ParachainStakingApi<Block, AccountId, Balance> for Runtime {
+		fn get_unclaimed_staking_rewards(account: AccountId) -> BalanceWrapper<Balance> {
+			let result = ParachainStaking::get_unclaimed_staking_rewards(&account);
+			BalanceWrapper{amount:result}
```
ebma, line 1530: Nit: indentation

runtime/amplitude/Cargo.toml
```diff
@@ -24,6 +24,7 @@ smallvec = "1.9.0"
 
 # Local
 runtime-common = { path = "../common", default-features = false }
+module-pallet-staking-rpc-runtime-api = { path = "../../pallets/parachain-staking/rpc/runtime-api", default-features = false }
```
ebma, line 27: Nit: let's move it to the other rpc api dependencies

runtime/pendulum/src/lib.rs
```diff
@@ -67,7 +67,7 @@ use xcm_config::{XcmConfig, XcmOriginToTransactDispatchOrigin};
 
 use orml_currencies::BasicCurrencyAdapter;
 use orml_traits::{currency::MutationHooks, parameter_type_with_key};
-
+pub use module_oracle_rpc_runtime_api::BalanceWrapper;
```
ebma, line 70: Does it need to be `pub`?

pallets/parachain-staking/rpc/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,31 @@
+//! Runtime API definition for Parachain Staking.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+use parity_scale_codec::{Codec, Decode, Encode, MaxEncodedLen};
+use frame_support::scale_info::TypeInfo;
+use frame_support::sp_runtime::Perquintill;
+use sp_std::fmt::Debug;
+use module_oracle_rpc_runtime_api::BalanceWrapper;
```
ebma, line 8: Maybe it's not a good idea to use the type defined in another rpc runtime api here. I'm not sure if we even still need something like the `BalanceWrapper`. Seems like [KILT does not](https://github.com/KILTprotocol/kilt-node/blob/ba088c2bd3a40f2983ed884e58f3a0046eb529df/runtime-api/staking/src/lib.rs#L41), so maybe we can also live without it?

pallets/parachain-staking/rpc/Cargo.toml
gianfra-t: True. In any case KILT does not have this rpc part, should we put as author pendulum?

res/amplitude-spec-raw.json
gianfra-t: My mistake, I thought they were needed every time we modified something like the rpc. I will roll back on them.

pallets/parachain-staking/rpc/Cargo.toml
ebma: Sure 👍

pallets/parachain-staking/rpc/runtime-api/src/lib.rs
gianfra-t: I had the understanding given [this comment](https://github.com/pendulum-chain/spacewalk/blob/main/pallets/oracle/rpc/runtime-api/src/lib.rs#L13) in oracle pallet in spacewalk, that it was needed if we ever wanted to set balance as u128. Do you think that the issue is outdated?

runtime/pendulum/src/lib.rs
gianfra-t: No, we can remove it.

prayagd:
@ebma should we already create ticket for this on the portal?

ebma:
@prayagd a ticket to employ the exposed functions? Sure, we can create that if it does not yet exist. Note that we need to redeploy the collator nodes on each network for this to take effect though, so merging this PR does not immediately unblock the work on the Portal.

pallets/parachain-staking/rpc/runtime-api/src/lib.rs
gianfra-t: I can confirm the the return value can be fetched. It simply returns `{amount: balance}` where balance is represented as a string as expected. 
The only caveat is that I couldn't test this using `polkadot.js` since I am having trouble calling the rpc (any custom rpc), but this is a different problem.

ebma: Alright, thanks for testing it 👍

ebma (approved):
Ready for merge once the conflicts are resolved

ebma:
CI is failing, let's fix that first

gianfra-t:
Now it should be good!

ebma (approved):

PR #359, opened 2023-11-29
333 extend the orml tokens pallet so users are able to mint and burn custom assets

### LINKED ISSUE #333: Extend the orml-tokens pallet so users are able to mint and burn custom assets
To complete the implementation of the new asset type, CurrencyID, within Pendulum, we require the ability for users and issuers to mint and burn custom assets. This ticket focuses on extending the orml-tokens pallet to facilitate the minting and burning of these custom assets, including CurrencyID tokens.

## Requirements:

Extend the orml-tokens pallet within the Pendulum repository to support the minting and burning of custom assets, specifically the CurrencyID tokens.

gianfra-t:
Closes #333.

Note: This is still a very basic implementation. More functionality needs to be defined first.

### General Overview
This pallet allows any user to "create" or "take" a `CurrencyId:Token(u64)`, and define the controlling accounts.

Once taken, the owner can execute extrinsics `mint` and `burn` which calls `orml_currencies` functions `deposit` and `withdraw` respectively (this in turn will interact with `orml_tokens` as defined in the issue)

The `StorageMap` will hold the information about which `CurrencyId:Token(u64)` has already been taken, and the accounts that are able to mint, burn and transfer ownership.

### Root Capabilities
Root can only transfer the ownership of the currency, which essentially allows to perform any operation.


gianfra-t:
Some more things we should define before moving forward. Some of them I can think of right now are:
- Do we need just the basic `mint` and `burn` functionality or we also need freezing assets, force transfer, etc.
- Do we want to allow for the destruction of a currency as it is done in `assets_pallet`? If so, how do we deal with token holders?

pallets/orml-tokens-extension/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+
+	#[pallet::pallet]
+	pub struct Pallet<T>(_);
+
+	// The pallet's dispatchable functions.
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Create and take ownership of one CurrencyId
+		///
+		/// The creator will have full control of this pallelt's functions
```
ebma, line 107:
```suggestion
		/// The creator will have full control of this pallet's functions
```

pallets/orml-tokens-extension/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+		/// regarding this currency
+		///
+		/// Parameters:
+		/// - `currency_id`: Currency id of the Token(u64) variant.
```
ebma, line 111: Let's maybe not mention the `Token(u64)` here as it's not explicitly mentioned/used anywhere in this pallet.

ebma (commented):
Looks good overall. We should probably add benchmarks but in my opinion we don't need to support freezing or anything fancy just yet.

pallets/orml-tokens-extension/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+sp-std = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.40", default-features = false}
+
+frame-benchmarking = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.40", default-features = false, optional = true }
+
+orml-currencies = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.40", default-features = false }
+orml-tokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.40", default-features = false }
+orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.40", default-features = false }
+
+# Spacewalk libraries
+spacewalk-primitives = { git = "https://github.com/pendulum-chain/spacewalk", default-features = false, rev = "d05b0015d15ca39cc780889bcc095335e9862a36"}
```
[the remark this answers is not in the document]
gianfra-t: Yess

pallets/orml-tokens-extension/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+			currency_id: CurrencyOf<T>,
+			new_admin: AccountIdOf<T>,
+			new_issuer: AccountIdOf<T>,
+		) -> DispatchResult {
+			let origin = ensure_signed(origin)?;
+
+			CurrencyData::<T>::try_mutate(currency_id.clone(), |maybe_details| {
+				let details = maybe_details.as_mut().ok_or(Error::<T>::NotCreated)?;
+
+				ensure!(origin == details.owner, Error::<T>::NoPermission);
```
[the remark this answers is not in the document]
gianfra-t: Are you thinking all of them? also minting ?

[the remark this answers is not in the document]
ebma: Yeah also minting and burning. Or do you think the root user should not be able to do that?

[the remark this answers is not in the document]
gianfra-t: I don't know! I think this is more a product definition. But I will add it, since I agree that at least at the beginning is better to have it.

runtime/amplitude/src/lib.rs
```diff
@@ -1117,11 +1120,7 @@ impl oracle::Config for Runtime {
 	type WeightInfo = oracle::SubstrateWeight<Runtime>;
 	type DataProvider = DataProviderImpl;
 	#[cfg(feature = "runtime-benchmarks")]
-	type DataFeedProvider = DataFeederBenchmark<
-		oracle::OracleKey,
-		oracle::TimestampedValue<UnsignedFixedPoint, Moment>,
-		Self::AccountId,
-	>;
+	type DataFeeder = MockDataFeeder<Self::AccountId,Moment>;
```
ebma, line 1123: Could we use the `MockDataFeeder` that is used in Spacewalk's oracle pallet altogether or is there an issue with re-using that?

gianfra-t: The problem is that it makes uses of some globally defines variables [COINS and LOCK](https://github.com/pendulum-chain/spacewalk/blob/13d25ae642cf43a65ebcb6028f5f0bc5b4e92160/pallets/oracle/src/testing_utils/mock_collector.rs#L105-L127)

ebma: But does it matter if we just import it from the crate itself, like [here](https://github.com/pendulum-chain/spacewalk/blob/13d25ae642cf43a65ebcb6028f5f0bc5b4e92160/testchain/runtime/src/lib.rs#L514)?

gianfra-t: Actually you are right! I'll remove the mock here.

ebma:
Is this still a draft? Or is it ready for review now @gianfra-t?

gianfra-t:
I believe it's ready. It does not have much functionality though but we decided to make destruction a different ticket as I recall.

pallets/orml-tokens-extension/Cargo.toml
```diff
@@ -0,0 +1,61 @@
+[package]
+authors = ["Pendulum Chain"]
+edition = "2021"
+name = "orml-tokens-extension"
```
ebma, line 4: Maybe we should give it a more descriptive name so that it also indicates the feature we are extending it with. How about we call it `orml-tokens-management-extension`?

pallets/orml-tokens-extension/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
+		/// Some currency was issued.
+		Mint { currency_id: CurrencyOf<T>, to: AccountIdOf<T>, amount: BalanceOf<T> },
+		/// Some currency was burned.
+		Burned { currency_id: CurrencyOf<T>, from: AccountIdOf<T>, amount: BalanceOf<T> },
+		/// Some currency class was created.
+		Created { currency_id: CurrencyOf<T>, creator: AccountIdOf<T>, owner: AccountIdOf<T> },
+		/// Some currency data was destroyed
+		Destroyed { currency_id: CurrencyOf<T> },
```
ebma, line 82: Maybe we should remove this event until we actually add the destruction of assets.

pallets/orml-tokens-extension/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+		///
+		/// The origin must be Signed and the sender must be the Admin of the currency `id`.
+		///
+		/// - `id`: The identifier of the currency to have some amount burned.
+		/// - `from`: The account to be debited.
+		/// - `amount`: The amount of the currency to be burned.
+		///
+		/// Emits `Burned` event when successful.
+		///
+		/// Weight: `O(1)`
```
ebma, line 196: Maybe we should remove all these 'Weight' comments now that we have proper benchmarks

pallets/orml-tokens-extension/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+			})
+		}
+
+		/// Change the Issuer and Admin.
+		///
+		/// Origin must be Signed and the sender should be the Owner of the currency.
+		///
+		/// - `currency_id`: Identifier of the currency.
+		/// - `issuer`: The new Issuer of this currency.
+		/// - `admin`: The new Admin of this currency.
```
ebma, line 294: Nit: wrong order and names of the parameters

pallets/orml-tokens-extension/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+
+		/// Weight information for the extrinsics in this module.
+		type WeightInfo: WeightInfo;
+
+		/// Type that allows for checking if currency type is ownable by users
+		type CurrencyIdChecker: CurrencyIdCheck<CurrencyId = CurrencyOf<Self>>;
+
+		/// TODO needs to be conditionaly compiled
+		#[cfg(feature = "runtime-benchmarks")]
+		type GetTestCurrency: Get<CurrencyOf<Self>>;
```
ebma, line 64: We don't necessarily need this config type at all. Instead of passing a currency just for testing and benchmarking purposes, we can either (in tests and benchmarks):
- get the native currency ID from teh `orml_currencies` pallet, like [here](https://github.com/pendulum-chain/pendulum/blob/10f3b7fa66995a0cfa93b26511209db47481e3c6/pallets/orml-currencies-allowance-extension/src/benchmarking.rs#L13)
- or just define the `CurrencyId` type as a simple `u64` and use numbers in the tests and benchmarks, see [here](https://github.com/pendulum-chain/pendulum/blob/10f3b7fa66995a0cfa93b26511209db47481e3c6/pallets/orml-currencies-allowance-extension/src/mock.rs#L36) and [here](https://github.com/pendulum-chain/pendulum/blob/10f3b7fa66995a0cfa93b26511209db47481e3c6/pallets/orml-currencies-allowance-extension/src/benchmarking.rs#L17)

And then we can also adjust the `CurrencyIdChecker` to always allow `u64` numbers in the mock.

gianfra-t: I actually had some troubles when using the `concrete` currency in the benchmarks. 
- Native currency I tried but throws an error when minting.
- Second solution would be fine, I tried to use the non-generic `CurrencyId:Token(u64)` on the benchmarks but couldn't make that work due to type mismatch, that's why I added the conditional config type. 
I didn't try with u64, I'll give it a try.

pallets/orml-currencies-allowance-extension/src/mock.rs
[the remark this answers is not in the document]
gianfra-t: It's a mistake. I must have confused mocks.

ebma (approved):
Great 👍 ready for merge IMO

PR #365, opened 2023-12-04
358 handle delegation of brz token to automation pallet in xcm config

gianfra-t:
A [custom](https://github.com/pendulum-chain/pendulum/pull/365/files#diff-cc6b60f53772abf50953b1259ee747096df19845aace835312ea8fe360d489a2R49) `MultiCurrencyAdapter` that is used as `AssetTransactor` is re-defined in runtime-common which then can be used to specify [on the xcm config of the runtimes](https://github.com/pendulum-chain/pendulum/pull/365/files#diff-91848e42611fbf16b8a970242187d57e75083383e231b2a458f8404b2d08e920R258) the `asset` and `beneficiary_location` that we want to "intercept", corresponding to the `ReserveAssetDeposited`'s `asset` an `DepositAsset`'s `beneficiary` instruction on the XCM message.

### Detailed Explanation
Any transfer from the `XTokens` pallet in moonbeam corresponding to the transfer of a native `XC-20` token, will trigger the following instructions:
`ReserveAssetDeposited, ClearOrigin, BuyExecution, DepositAsset`

See for instance [this event](https://moonbeam.subscan.io/extrinsic/5024347-5?event=5024347-15) which corresponds to the transfer of a local XC 20 with account key `0x931715fee2d06333043d11f658c8ce934ac61d0c`.
The corresponding XCM message for this event can be found [here](https://moonbeam.subscan.io/xcm_message/polkadot-552181141a87215dc58f8b766e7ae54d8bb01fab) where the `ReserveAssetDeposited` instructions reflects the multilocation of the XC-20 asset, and `DepositAsset` that of the destination defined in the `XTokens::transfer`.

The call to the [deposit_asset](https://github.com/pendulum-chain/pendulum/pull/365/files#diff-cc6b60f53772abf50953b1259ee747096df19845aace835312ea8fe360d489a2R96) function will happen with the asset defined in `ReserveAssetDeposited` and the beneficiary location being that of the automation pallet.


runtime/amplitude/src/xcm_config.rs
```diff
@@ -246,8 +266,77 @@ impl ShouldExecute for DenyReserveTransferToRelayChain {
...
+		vec![
+			MatcherPair::new(
+				Box::new(ReserveAssetDepositedMatcher1),
+				Box::new(DepositAssetMatcher1),
+			),
+			// Additional matcher pairs to be defined in the future
+		]
+	}
+	fn get_incoming_parachain_id() -> u32 {
+		MOONRIVER_PARA_ID
```
b-yap, line 329: I see you've added a comment in line 325, about additional match pairs.
But  `get_incoming_parachain_id()` is specific to `MOONRIVER_PARA_ID`. Does it mean only tokens from _**Moonriver**_ are supported? 

If that's the intention, maybe rename the struct to `MoonRiverMatcherForAmplitude` or something, to make it obvious on the first read.

runtime/common/src/custom_xcm_barrier.rs
```diff
@@ -0,0 +1,121 @@
...
+		_max_weight: XCMWeight,
+		_weight_credit: &mut XCMWeight,
+	) -> Result<(), ()> {
+		log::info!(
+			target: "xcm::barriers",
+			"AllowUnpaidExecutionFromCustom origin: {:?}, instructions: {:?}, max_weight: {:?}, weight_credit: {:?}",
+			origin, instructions, _max_weight, _weight_credit,
+		);
+		log::info!("origin {:?}", origin);
+		let incoming_parachain_id = V::get_incoming_parachain_id();
```
b-yap, line 72: I know this ticket is only for BRZ. But in the future, are we also thinking about supporting other tokens? From other parachains?

  For Amplitude, _`V`_ here is `MatcherConfigAmplitude`, which the func `get_incoming_parachain_id()` returns **_`MOONRIVER_PARA_ID`_**  all the time.

It seems very restrictive. 

@ebma @gianfra-t Please ignore this comment if that's what we're  aiming for.

ebma (commented):
It looks good overall, great job @gianfra-t 👍 
I think it would be better if we don't add the same barrier on all runtimes though. For now I would only add it to the pendulum runtime and revert the changes for Foucoco and Pendulum. The reason is that the BRZ location only is correct like this on Pendulum. For testing purposes we will probably add a very similar barrier to Foucoco as well (but defining a different contract location instead of BRZ because we need to be the issuer of that contract for testing), so it's great that the barrier is implemented in the `common` directory and re-usable. Adding the changes for Foucoco should be part of a different ticket once we know the details though.

Thus, I'd suggest we only focus on Pendulum for now and also change the added integration test to only target Pendulum and not the other runtimes.

runtime/common/src/custom_xcm_barrier.rs
```diff
@@ -0,0 +1,121 @@
...
+									return Err(())
+								}
+							}
+						}
+					}
+				}
+			}
+		}
+
+		ensure!(T::contains(origin), ());
```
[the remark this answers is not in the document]
gianfra-t: Yes probably since origin will be also cleared.

TorstenStueber (commented):
Nice PR. I have some general remarks.

* Is the Barrier really the right point to execute side effects? Just the name of the trait `ShouldExecute` indicates that this is just a pure function that returns Yes or No. Do we have examples from other parachains?
* The xcm message can fail at a later point and we execute the side effect anyway (or am I wrong here?)
* The whole trait `MatcherConfig` feels overly generic to me. Let's please always just implement the minimum required and add features later if we need them, particularly:
  * why a fee system?
  * why a vector of matcher pairs?
  * why does is the `MatcherPair` a struct of functions instead of just a struct with an expected multilocation and, well, the `DepositAssetMatcher` can probably not be simplified that much

runtime/pendulum/src/xcm_config.rs
```diff
@@ -262,7 +268,83 @@ impl ShouldExecute for DenyReserveTransferToRelayChain {
...
+	fn matches<'a>(
+		&self,
+		assets: &'a MultiAssetFilter,
+		beneficiary: &'a MultiLocation,
+	) -> Option<(u8, &'a [u8])> {
+		if let (
+			Wild(AllCounted(1)),
+			MultiLocation {
+				parents: 0,
+				interior: X2(PalletInstance(99), GeneralKey { length, data }),
```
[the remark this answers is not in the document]
gianfra-t: Yes!

gianfra-t:
To address some of the remarks (as far as I can!):

> Is the Barrier really the right point to execute side effects? Just the name of the trait ShouldExecute indicates that this is just a pure function that returns Yes or No. Do we have examples from other parachains?

At first I actually chose the `FungiblesAdapter` to hold our custom logic, but at that point in the execution we don't have the origin and the `ReserveAssetDeposited` instruction which we need to confirm that the asset is the BRZ XC-20 token. I agree that the barrier is not the most elegant solution, but is the only place I could think at the moment that had the full set of instruction of the XCM message, and that is simple enough to define custom logic. I could not find any example in other parachains with a similar case, and to be honest don't know how to approach that search.  

> The xcm message can fail at a later point and we execute the side effect anyway (or am I wrong here?)

I don't think it can fail really, since it is not going to be executed by the `xcm_executor` anyway. Once we match the special case in this custom barrier, you can see that actually the barrier [denies](https://github.com/pendulum-chain/pendulum/pull/365/files#diff-e3c102617e9fd24e83e4146117434fb789fdcbcd1582004d9288e31d2992c9b0R124) the message, because we don't want it processed in the "normal route", and we have already processed the call to the automation pallet. 

> The whole trait MatcherConfig feels overly generic to me. Let's please always just implement the minimum required and add features later if we need them 

This is true, it is not the "simplest" solution. The idea of using a vector is to allow for easily handling a new XC-20 token like BRZ that will go the same route (to the automation pallet). If this is not needed then we can simplify a bit. But fees was something that was discussed before and since the message will never reach the `BuyExecution` part, we needed some way to define how to handle fees.

TorstenStueber:
Wouldn't the `AssetTransactor` config of `xcm-executor` be the right place? This is exactly the place where the assets are supposed to be moved into our special multilocation. The `deposit_asset` function of `AssetTransactor` will be called in `DepositAsset` – this is where the funds in the holding accounts are moved to the destination multilocation.

I think that this is better and more flexible then reading the complete xcm message and trying to interpret what it does. Instead let the xcm executor execute the whole message and insert our special code only at the place where the real transaction happens.

> I don't think it can fail really, since it is not going to be executed by the xcm_executor anyway. Once we match the special case in this custom barrier, you can see that actually the barrier [denies](https://github.com/pendulum-chain/pendulum/pull/365/files#diff-e3c102617e9fd24e83e4146117434fb789fdcbcd1582004d9288e31d2992c9b0R124) the message, because we don't want it processed in the "normal route", and we have already processed the call to the automation pallet.

I missed that and I would definitely advise against this. We put too many assumptions here on how messages sent by other parachains will look like in the future. I think we should definitely not skip the execution of the xcm executor pallet, particularly all the safety mechanisms built into it (and, e.g., fee payments, etc.)

> But fees was something that was discussed before and since the message will never reach the BuyExecution part, we needed some way to define how to handle fees.

If this is not a definition in the issue description, then we should use the normal fee mechanism of the xcm-executor and not start to build a custom solution here. The decision and discussion should happen in the issue instead.

gianfra-t:
You are right that the `AssetTransactor`'s `deposit_asset` is the most logical place, something like [this](https://github.com/pendulum-chain/pendulum/pull/365/commits/282e08d50487773b7caa66098c61c6e47fe0a307#diff-fcdbf97066e880f9f60468982f9aad27a5ba97e7ad6c6c6370db1af0b0c588daR73). But the problem with that solution I see is that we don't have the `ReserveAssetDeposited` or more precisely it's `assets`, which are the ones which were withdrawn in moonbeam, since the instruction `DepositAsset` is the one that calls the `deposit_asset` of the transactor, with it's parameters that look like:
```
DepositAsset { assets: Wild(AllCounted(1)), beneficiary: MultiLocation { parents: 0, interior: X2(PalletInstance(99), GeneralKey { length ....
```
In this case, how could we prevent that another XC-20 token (BRZ_2) calls `transfer` with the same destination? The automation pallet would execute anyway.

> Instead let the xcm executor execute the whole message and insert our special code only at the place where the real transaction happens.

This I agree until the fee extraction, after that (the deposit instruction left) will fail anyway since we will not map this special destination into a token.

TorstenStueber:
Yes, it is exactly the `_what` part of `deposit_asset`. I think this is an important discussion: I consider it not important where the token is coming from. I see our special multilocation as a black hole that would automatically trigger our automation logic. Any token that arrives in that black hole – from wherever – would then be forwarded to our automation logic.

Am I naive here? Could this be exploited? If that could be exploited, that means that another parachain would be able to send us an XCM message that we execute blindly that contains the `DepositAsset` instruction in a "malicious way". Is that even a scenario? Could such a hostile chain exist that just sends malicious `DepositAsset` instructions to us freely? I think that this is exactly why we have the `Barrier`, to check that xcm messages are well formed.

gianfra-t:
Then assuming we trust moonbeam and the other connected parachains which is probably a good assumption, and that we do not check for the parachain origin,  the only dangerous scenario I see  is another parachain that also implements pallet contracts or similar, where some user would be able to replicate this specific set of instructions. 

> But even if we can ensure that the xcm message originates at Moonbeam and is well formed: we do not know whether Moonbeam actually locked a BRZ token or whether Moonbeam just sent this xcm message to us without any BRZ tokens being sent via XCM. 

In this case I think we do because it is reserved asset transfer as the ones [here](https://moonbeam.subscan.io/xcm_message/polkadot-552181141a87215dc58f8b766e7ae54d8bb01fab) right?

TorstenStueber:
> Then assuming we trust moonbeam and the other connected parachains which is probably a good assumption, and that we do not check for the parachain origin, the only dangerous scenario I see is another parachain that also implements pallet contracts or similar, where some user would be able to replicate this specific set of instructions.

What do you mean, can you elaborate? I don't think that users are themselves able to generate these instructions. Users are not able to send arbitrary XCM messages to us from another parachain.

> In this case I think we do because it is reserved asset transfer as the ones [here](https://moonbeam.subscan.io/xcm_message/polkadot-552181141a87215dc58f8b766e7ae54d8bb01fab) right?

Sure, this is a reserved asset transfer. But just in principle: if Moonbeam implements malicious logic, they could just send us an XCM message that looks exactly like the one that results from a reserve asset transfer.

gianfra-t:
> I don't think that users are themselves able to generate these instructions. Users are not able to send arbitrary XCM messages to us from another parachain.
For sure, I was thinking more about a parachain having a pallet contracts or equivalent that allowed their users to create tokens (like the XC-20) and send to a desired destination. I honestly don't know how it would look like, and it may be too far fetched.

I also wasn't aware of the checking in `ReserveAssetDeposited`. So we can say that by checking the `_what` we are already checking the origin. 
In any case I will rollback to using a custom `AssetTransactor` as the place for our custom instructions.

ebma:
>  I don't think that users are themselves able to generate these instructions. Users are not able to send arbitrary XCM messages to us from another parachain.

Maybe it misses the point and you meant something else, but users actually are able to send arbitrary XCM messages to us from other parachains using the [send()](https://github.com/paritytech/polkadot-sdk/blob/8efaabd6c1c0cecc8c0d5ec94ac7877407223845/polkadot/xcm/pallet-xcm/src/lib.rs#L890-L897) extrinsic of the xcm-pallet @TorstenStueber.

gianfra-t:
> users actually are able to send arbitrary XCM messages to us from other parachains using the [send()](https://github.com/paritytech/polkadot-sdk/blob/8efaabd6c1c0cecc8c0d5ec94ac7877407223845/polkadot/xcm/pallet-xcm/src/lib.rs#L890-L897) extrinsic of the xcm-pallet

But I think one can control that with `SendXcmOrigin` trait.

TorstenStueber:
@ebma I was not aware. I assume that in this case the origin MultiLocation will be different and not just `{ parent: 1, interior: X1(Parachain(...)) }`, correct?

ebma:
> I assume that in this case the origin MultiLocation will be different and not just { parent: 1, interior: X1(Parachain(...)) }, correct?

No, I think it will be the correct `MultiLocation` of the sending chain. For example, you could use the `polkadotXcm::send()` extrinsic on Moonbeam to build an XCM transaction with the same instructions that would be created as a result of a call to  `xtokens::transfer()` pallet. I don't think we would be able to tell the difference on our receiving chain.

gianfra-t:
@ebma I think you are correct that anyone can send a `DepositAsset` instruction with the form we are intercepting, but I don't think it will execute.
If we look at the `DepositAsset` code, it first attempts to [retrieve](https://github.com/paritytech/polkadot/blob/52209dcfe546ff39cc031b92d64e787e7e8264d4/xcm/xcm-executor/src/lib.rs#L619) the assets from the holding. And the assets may only arrive before when either `WithdrawAsset`, `ReserveAssetDeposited` or `ReceiveTeleportedAsset` execute [subsume_asset](https://github.com/paritytech/polkadot/blob/52209dcfe546ff39cc031b92d64e787e7e8264d4/xcm/xcm-executor/src/lib.rs#L497). So if someone sends that instruction there will be no assets in the holding, and `deposit_asset` will not be called. 
Otherwise this would be a problem for all other xcm transactions.

TorstenStueber:
@ebma @gianfra-t If everyone can send an xcm message to us from Moonbeam and we are not able to tell the difference, then this would be a clear flaw: the whole purpose of a reserve asset transfer is that such an xcm message only arrives at Pendulum when Moonbeam locked an amount of tokens before – like for every bridge. When we execute the xcm message, then we mint an equivalent amount of wrapped tokens. This should of course never happen when Moonbeam did not lock tokens before – otherwise this would be a flawed bridge design.

TorstenStueber:
I just checked this:
* the `send` extrinsic will execute the `send_xcm` function in pallet-xcm where `interior` is `X1(AccountId32({network: Polkadot, id: <the account id of the caller>}))` (see [here](https://github.com/moonbeam-foundation/polkadot-sdk/blob/647bc10a95f792d2bb30414860be91411cb21d43/polkadot/xcm/pallet-xcm/src/lib.rs#L781C5-L781C5))
* then `send_xcm` will add a new instruction `DescendOrigin(interior)` as the first instruction of the xcm message (see [here](https://github.com/moonbeam-foundation/polkadot-sdk/blob/647bc10a95f792d2bb30414860be91411cb21d43/polkadot/xcm/pallet-xcm/src/lib.rs#L1493-L1494))
* on our side the `DescendOrigin` instruction will then be executed first and ensure that the origin register of the xcm VM will be changed from `{parent: 1, X1(Parachain(...)}` to `{parent: 1, X2(Parachain(...), AccountId32(...))}` (see [here](https://github.com/paritytech/polkadot/blob/95fe4c8862810bffd68343231a517e62689c05c0/xcm/xcm-executor/src/lib.rs#L591-L597))
* when we then execute the next xcm instruction `ReserveAssetDeposited`, our xcm VM will check whether the origin register is just the parachain where the asset that is sent to us originates (see [here](https://github.com/paritytech/polkadot/blob/95fe4c8862810bffd68343231a517e62689c05c0/xcm/xcm-executor/src/lib.rs#L489-L492))
* which will not hold and the result is that our VM execution will fail with `XcmError::UntrustedReserveLocation`

One more reason why we should not skip the normal VM execution (as you did originally @gianfra-t).

ebma:
> - when we then execute the next xcm instruction ReserveAssetDeposited, our xcm VM will check whether the origin register is just the parachain w[here](https://github.com/paritytech/polkadot/blob/95fe4c8862810bffd68343231a517e62689c05c0/xcm/xcm-executor/src/lib.rs#L489-L492) the asset that is sent to us originates (see here)
> - which will not hold and the result is that our VM execution will fail with XcmError::UntrustedReserveLocation

If I understand correctly this does not hold because the origin of this `polkadotXcm::send()` message will be a `Junction` with a specific account while when using the `xtokens` pallet the origin will be a `Junction` that has the parachain itself as the origin?

TorstenStueber:
I did not check what the xtokens pallet does. I [only checked](https://www.notion.so/satoshipay/23-12-07-Deposit-Under-the-Hood-a18edad841054363a64b7baca3ad87bd) what pallet-xcm does when executing the `reserve_transfer_assets` extrinsic.

I assume that xtokens works similarly but it might depends on the concrete extrinsic. What extrinsic do you have in mind?

ebma:
I was just curious as to why I wouldn't be able to execute a reserve transfer of assets with `polkadotXcm::send()` and I assume the answer is that as you pointed out [here](https://github.com/moonbeam-foundation/polkadot-sdk/blob/647bc10a95f792d2bb30414860be91411cb21d43/polkadot/xcm/pallet-xcm/src/lib.rs#L774), the origin will be set to `{parent: 1, X2(Parachain(...), AccountId32(...))}` but for the reserve transfer to work when being processed on the target chain, the origin would need to be just `{parent: 1, X1(Parachain(...)}`? Or am I wrong? 

I checked the xtokens pallet and when doing a `transfer()` eventually it will create an XCM message similar to what you wrote down in the Notion page, see [here](https://github.com/open-web3-stack/open-runtime-module-library/blob/84c58de7820d49acf81e7e4b6a5e28f8572bef95/xtokens/src/lib.rs#L690-L705).

TorstenStueber:
I find it confusing that there is pallet-xcm and xtokens, seems like there is a big overlap in their purpose.

So it seems like the following two extrinsics basically have the same outcome:
* `pallet-xcm` -> `limited_reserve_transfer_assets`
* `xtokens` -> `transfer_multiasset`


Both of these extrinsics will ultimately call the function `execute_xcm_in_credit` in the `xcm-executor` pallet and will provide the following parameters:
* `origin`: the multilocation `{parents: 0, interior: X1(AccountId32 { ... }) }`
* `message`: 
    ```
    TransferReserveAsset {
      assets: ... // the vector of assets to be sent,
      dest: ... // the multilocation of the target parachain,
      xcm: Xcm(vec![
        BuyExecution {
          fee: ... // the fee multiasset reanchored to the target parachain
          weight_limit: ... // some weight limit
        },
        DepositAsset {
          assets: Wild(AllCounted(/* number of assets to send */)), 
          beneficiary: ... // the recipient multilocation
        },
      ]),
    }
    ```

runtime/pendulum/src/xcm_config.rs
```diff
@@ -50,9 +53,6 @@ parameter_types! {
 		X2(GlobalConsensus(RelayNetwork::get()), Parachain(ParachainInfo::parachain_id().into()));
 }
 
-/// Type for specifying how a `MultiLocation` can be converted into an `AccountId`. This is used
-/// when determining ownership of accounts for asset transacting and when attempting to use XCM
-/// `Transact` in order to determine the dispatch Origin.
 pub type LocationToAccountId = (
```
ebma, line 56: Can't we keep the doc comment?

runtime/pendulum/src/xcm_config.rs
```diff
@@ -262,6 +252,54 @@ impl ShouldExecute for DenyReserveTransferToRelayChain {
...
+
+		match asset {
+			MultiAsset { id: AssetId::Concrete(loc), fun: Fungibility::Fungible(amount_deposited) } if loc == &expected_multiloc =>
+				return Some(*amount_deposited),
+			_ => return None,
+		}
+	}
+
+	// TODO modify with automation's pallet instance
+	fn matches_beneficiary(beneficiary_location: &MultiLocation) -> Option<(u8, [u8;32])> {
```
ebma, line 270: Maybe we should create an extra struct for the return type? It's not obvious why we return a tuple `(u8, [u8;32])` here but if we group it under some struct with a descriptive name it would be easier to understand.

runtime/common/src/custom_transactor.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+pub trait AutomationPalletConfig {
+	fn matches_asset(asset: &MultiAsset) -> Option<u128>;
+	fn matches_beneficiary(beneficiary_location: &MultiLocation) -> Option<(u8, [u8;32])>;
+	fn callback(length: u8, data: [u8;32], amount: u128) -> Result;
+}
+
+#[allow(clippy::type_complexity)]
+pub struct CustomMultiCurrencyAdapter<
```
ebma, line 49: This draws heavy inspiration from ORMLs [MultiCurrencyAdapter](https://github.com/open-web3-stack/open-runtime-module-library/blob/84c58de7820d49acf81e7e4b6a5e28f8572bef95/xcm-support/src/currency_adapter.rs#L105) which is good because there is obviously no need to reinvent the wheel here. I'm just wondering if it wouldn't be better to change this struct so that it takes two traits, 1) the existing `MultiCurrencyAdapter` and the `AutomationPalletConfigT`. And then we make the `deposit/withdraw/transfer_asset()` functions just call the methods on the `MultiCurrencyAdapter` that was passed to us but for the `deposit_asset()` we add the extra automation pallet check before calling the 'parent' function.
Would this work, what do you think @gianfra-t? This way we can deduplicate some code and profit from upstream changes made to the `MultiCurrencyAdapter` of orml, if any.

gianfra-t: That is a very good point, I can't believe I didn't think on that. We can just require that trait 1) just implements `deposit` and `withdraw` so we can later switch it for something else.

gianfra-t: I changed it so the "WrappedTransactor" (`MultiCurrencyAdapter` in this case) has to implement xcm's `TransactAsset`, which is anyways what `AssetTransactor` must implement so I guess this is as generic as we can go.

ebma (approved):
Nice, looks good to me now 👍 

What do you think @TorstenStueber?

TorstenStueber:
Looks good so far. I want to check a few things and will give my final opinion.

TorstenStueber (approved):
Great job @gianfra-t! Thanks.

PR #367, opened 2023-12-05
Register Moonbase DEV on Moonbeam Foucoco testnet

### LINKED ISSUE #366: Register Moonbase Alpha's DEV token on Foucoco
In order to finish the testnet integration of #318, we need to transfer tokens between our test chain and Moonbase Alpha. For this, we'll add the native token of Moonbase Alpha (DEV) to the XCM config of our Foucoco runtime. This is described [here](https://docs.moonbeam.network/builders/interoperability/xcm/xc-registration/assets/#register-moonbeam-native-assets).

## TODO
- [x] Register the DEV token in the Foucoco runtime
- [x] #369
- [x] #371

bogdanS98:
Closes #366.

### DIFF
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -206,7 +206,7 @@ pub mod parachains {
 	
 			parachain_asset_location!(
 				DEV,
-				X2(Parachain(PARA_ID), PalletInstance(BALANCES_PALLET_INDEX),)
+				X2(Parachain(PARA_ID), PalletInstance(BALANCES_PALLET_INDEX))
 			);
 		}
 	}
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -99,6 +99,7 @@ pub mod parachains {
 	}
 
 	/// Creates a location for the given asset in this format: `fn <asset_name>_location() -> MultiLocation`
+	#[macro_export]
 	macro_rules! parachain_asset_location {
 		// Also declares a constant variable <asset_name>_ASSET_ID with <asset_value>.
 		// This assumes that the following constant variables exist:
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -99,7 +99,6 @@ pub mod parachains {
 	}
 
 	/// Creates a location for the given asset in this format: `fn <asset_name>_location() -> MultiLocation`
-	#[macro_export]
 	macro_rules! parachain_asset_location {
 		// Also declares a constant variable <asset_name>_ASSET_ID with <asset_value>.
 		// This assumes that the following constant variables exist:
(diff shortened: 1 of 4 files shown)


runtime/foucoco/src/assets.rs
```diff
@@ -35,3 +36,21 @@ pub mod native_locations {
 		}
 	}
 }
+
+pub mod moonbase_alpha_relay {
+	pub mod moonbase {
```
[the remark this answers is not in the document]
ebma: Maybe it's also best to move this to the runtime/common location to where the polkadot and kusama stuff is defined. While it's not shared across all runtimes and only used for Foucoco, structurally it makes most sense to group it with the others.

runtime/foucoco/src/assets.rs
```diff
@@ -35,3 +36,21 @@ pub mod native_locations {
...
+		use runtime_common::parachain_asset_location;
+		use xcm::latest::{
+			Junction::{PalletInstance, Parachain},
+			Junctions::X2,
+		};
+
+		pub const PARA_ID: u32 = 1000;
+		pub const BALANCES_PALLET_INDEX: u8 = 110;
```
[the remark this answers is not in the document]
bogdanS98: Yes, that's right, I copied that value by mistake

runtime/common/src/lib.rs
```diff
@@ -193,6 +193,23 @@ pub mod parachains {
...
+				Junction::{PalletInstance, Parachain},
+				Junctions::X2,
+			};
+	
+			pub const PARA_ID: u32 = 1000;
+			pub const BALANCES_PALLET_INDEX: u8 = 3;
+	
+			parachain_asset_location!(
+				DEV,
+				X2(Parachain(PARA_ID), PalletInstance(BALANCES_PALLET_INDEX),)
```
ebma, line 209: Do we need the comma? 🤔 

```suggestion
				X2(Parachain(PARA_ID), PalletInstance(BALANCES_PALLET_INDEX))
```

runtime/foucoco/src/assets.rs
```diff
@@ -3,6 +3,7 @@
 pub mod xcm_assets {
 	use runtime_common::create_xcm_id;
 	create_xcm_id!(RELAY, 0);
+	create_xcm_id!(MOONBASE_DEV, 2);
```
ebma, line 6: Should we just use `1`? Or is there a reason you chose `2`?

ebma (approved):
Looks good 👍

runtime/common/src/lib.rs
bogdanS98: No, we can delete it 👍🏼

runtime/foucoco/src/assets.rs
bogdanS98: I chose `2` because I thought that if we need to register some new asset on Rococo Foucoco in the future to have that asset to be `1`

bogdanS98: Or we could just use `1`, let me know what you think

ebma: Let's just go with `1`, if that's not yet assigned to something else. (And as far as I know it's not)

ebma (approved):
Looks good I think 👍

PR #375, opened 2023-12-11
Copy Foucoco's chain extension impl to Amplitude and Pendulum

### LINKED ISSUE #356: Add ERC20 and oracle chain extension to Amplitude and Pendulum
The chain extension developed for Foucoco that contains the code used by the ERC20 wrapper contracts and the oracle wrapper contract should also be added to Amplitude and Pendulum.

It might makes sense to deduplicate code and move the chain extension into runtime/common, however, we should carefully check what effects this can have and whether the code is compatible with all three runtimes.

b-yap:
Closes #356 . 

As mentioned in the [last comment](https://github.com/pendulum-chain/pendulum/issues/356#issuecomment-1831654713), this is just copying the chain extension in Foucoco to Amplitude and Pendulum.


gianfra-t (commented):
Looks good!. I've got just one question, in the pendulum [pallet_contracts config](https://github.com/pendulum-chain/pendulum/blob/e07140a8cac4ee66327fe9945bb6ecffd2346d3b/runtime/pendulum/src/lib.rs#L864) shouldn't we put `type ChainExtension = Psp22Extension;` like in amplitude and foucoco?

runtime/foucoco/Cargo.toml
```diff
@@ -192,7 +192,6 @@ std = [
     "pallet-xcm/std",
     "parachain-info/std",
     "parachain-staking/std",
-    "orml-currencies-allowance-extension/std",
```
ebma, line 195: Why are we removing this?

ebma (changes requested):
There are some formatting issues and @gianfra-t is right that the declaration for using the chain extension is missing on Pendulum. Rest looks good

runtime/foucoco/Cargo.toml
b-yap: This was added twice.
It's a duplicate of [this line](https://github.com/pendulum-chain/pendulum/blob/58d806850b912e22905c444d99db7a9b0757df98/runtime/foucoco/Cargo.toml#L240).

b-yap:
tested using the latest compiled contract of [ERC20Wrapper](https://github.com/pendulum-chain/pendulum-ink-wrapper/tree/0cf83d833457c4172ce80fecb243033bf94b1aea) 
Screen Shot 2023-12-19 at 11 23 16 PM
Screen Shot 2023-12-19 at 11 23 26 PM



Screen Shot 2023-12-19 at 11 30 38 PM

gianfra-t (approved):
I think this is good! especially since you tested the wrapped ERC20.

PR #377, opened 2023-12-11
376 xcm transfer of native currency does not work

### LINKED ISSUE #376: XCM transfer of native currency does not work
At the moment, the XCM transfer of our native currency does not work. One of (possibly many) reasons is that the `AssetTransactor` of the `xcm_executor` pallet config only supports interacting with the `Tokens` pallet right now, but we need to extend it such that it either uses the `Balances` pallet for our native currency or the `Tokens` pallet for all else.

There might be other issues with the XCM config though, which have to be identified when solving this issue.

ebma:
Uses the `Currencies` pallet to unify the handling of the native and other tokens in one handler. The `Currencies` pallet does not implement the [TransactAsset](https://github.com/paritytech/polkadot-sdk/blob/d43d2fe23b12355eed4afa6dd3e971a248269262/polkadot/xcm/xcm-executor/src/traits/transact_asset.rs#L30) trait which is required. However, orml provides an [adapter](https://github.com/open-web3-stack/open-runtime-module-library/blob/5c96d6bf12c5fc2aeac359f2297453ba63140927/xcm-support/src/currency_adapter.rs#L105) in order to use `Currencies` anyways. The implementation of our adapter is adopted from [Acala](https://github.com/AcalaNetwork/Acala/blob/49ae33ffcc98fe7b44311152f32b0fe98671ea9f/runtime/mandala/src/xcm_config.rs#L262) and [Interlay](https://github.com/interlay/interbtc/blob/786ef7e0075b980124036b9d9a8d78152d42290d/parachain/runtime/interlay/src/xcm_config.rs#L292). 

The configuration is so that if something fails when depositing the asset into the account of a user, the received asset is transferred to the treasury instead of getting lost. 

### UnknownTokens pallet
**Note** that I decided not to add the [unknown-tokens](https://github.com/open-web3-stack/open-runtime-module-library/tree/master/unknown-tokens) pallet. I added it at first (see some of the commits) because it is used by Acala and Interlay. The pallet has 0 documentation but from what I understand it could be used to allow receiving assets that are not registered on our chain. These assets are put into an extra register that is managed by the unknown-tokens pallet. They don't belong to a user but to a location and can only be deposited and withdrawn, not transferred. I assume that this implementation allows for use cases like transferring assets not known to Pendulum transitively, like a scenario where Parachain A wants to send an asset to Parachain B but they don't have direct channel and do a multi-hop transfer over Pendulum. Since this extends the functionality of our chain and is not part of this particular issue, I decided not to add this behaviour.

Closes #376.


TorstenStueber (commented):
@ebma Could we not replace `Tokens` with `Currencies` and that would then work again? This would be a little simpler.

ebma:
That's a great idea, thanks @TorstenStueber. I'll try to incorporate something like that.

gianfra-t (approved):
Seems all good.
By the way, do you know the impact of the [MultiCurrencyAdapter](https://github.com/open-web3-stack/open-runtime-module-library/blob/5c96d6bf12c5fc2aeac359f2297453ba63140927/xcm-support/src/currency_adapter.rs#L105) not implementing the `check_in` functions and the `internal_transfer_asset` that are on the previously used [FungiblesAdapter](FungiblesMutateAdapter)?

bogdanS98:
Looks good to me 👍🏼. Regarding your discussion @ebma @gianfra-t, indeed `check-in ` only seems to be used when teleporting and seems like `internal_transfer_asset` is used only if implemented. See `transfer_asset` [here](https://github.com/paritytech/polkadot-sdk/blob/d43d2fe23b12355eed4afa6dd3e971a248269262/polkadot/xcm/xcm-executor/src/traits/transact_asset.rs#L132-L152)

bogdanS98 (approved):

ebma:
> seems like internal_transfer_asset is used only if implemented. Usually, transfer_asset is called instead. See transfer_asset [here](https://github.com/paritytech/polkadot-sdk/blob/d43d2fe23b12355eed4afa6dd3e971a248269262/polkadot/xcm/xcm-executor/src/traits/transact_asset.rs#L132-L152)

Ah nice find, thanks @bogdanS98 👍

TorstenStueber:
Thanks @ebma, I agree that we should for now not add support for unknown tokens.

TorstenStueber (approved):
Looks all good to me.

PR #382, opened 2023-12-14
Update to 0.9.42

gianfra-t:

Closes [tasks/#158](https://github.com/pendulum-chain/tasks/issues/158)
 
 ### Changes in node/service.rs
 These changes where adopted from other chains like [equilibrium's](https://github.com/eq-lab/equilibrium/blob/polkadot-v0.9.42/node/src/service/mod.rs) which also uses `v0.9.42`.
 
 ### Updates to dependencies
 Spacewalk -> branch [upgrade-v0.9.42](https://github.com/pendulum-chain/spacewalk/tree/upgrade-v0.9.42)
 Substrate-stellar-sdk -> branch [polkadot-v0.9.42](https://github.com/pendulum-chain/substrate-stellar-sdk/tree/polkadot-v0.9.42)
 Dia-Oracle (oracle-pallet) -> branch [polkadot-v0.9.42](https://github.com/pendulum-chain/oracle-pallet/blob/polkadot-v0.9.42/pallets/dia-oracle/src/lib.rs)
 Pendulum/Bifrost -> branch [upgrade-v0.9.42](https://github.com/pendulum-chain/bifrost/tree/upgrade-v0.9.42). 
 
 ### Changes in runtime
 
 #### General
 New `TokenError` variants were added to `ChainExtensionError` enum and `From<>` functions. 
 
 #### Pallet Balances
 In all configs we must add the following new types (`FreezeIdentifier`, `MaxFreezes`, `MaxHolds` and `HoldIdentifier). Since we currently do not use any of these functionality, we can pass the default configuration. We need to adjust if we require more than one freeze or hold.
 ```
 type FreezeIdentifier = ();
  type MaxFreezes = ();
  type MaxHolds = ConstU32<1>;
  type HoldIdentifier = RuntimeHoldReason;
 ```
 
 Also in `pallet_balances` the `DustRemoval` trait bound has changed. Before it was required `type DustRemoval: OnUnbalanced<NegativeImbalance<Self, I>>;` 
Passing type `Treasure` is not supported anymore since the bound is now `type DustRemoval: OnUnbalanced<CreditOf<Self, I>>;` which is not implemented by the `Treasury` pallet, so we must define a new (link) implementation. This is adopted from [ComposableFi](https://github.com/ComposableFi/composable/blob/9a9846a7723b28c8d90fa8350d139a5374056fa4/code/parachain/runtime/picasso/src/assets.rs#L9) yet the functionality is the same which will transfer to the Treasury's account the unbalanced currency.

#### Pallet Contracts
- Removed `DeletionWeightLimit`, `DeletionQueueDepth` since are not required anymore.
- Adds `DefaultDepositLimit`, which is the fallback value to limit the storage deposit if it's not being set by the caller. We need to properly define this value.

#### Pallet Collective
 - Must add `MaxProposalWeight`. We use 50 percent of the max block weight as it is defined in [Frontier](https://github.com/paritytech/frontier-parachain-template/blob/ac10a6a534ea3a6e8f3a433f6ec1e08a7c8b76db/pallets/motion/src/mock.rs#L78).
 
 #### Pallet XCM
 - Adds `AdminOrigin` type, which is set to root. This is only used for privileged function about xcm version management.
 
 #### Farming Pallet
 Because we update our [bifrost fork](https://github.com/pendulum-chain/bifrost/tree/upgrade-v0.9.42) to `0.9.42`, the farming pallet now allows for a "boost reward" functionality, which can also be managed in conjunction with a new pallet `VeMinting`. This is refelected into 4 new config types for `farming` pallet: `FarmingBoost`, `VeMinting`, `BlockNumberToBalance`, `WhitelistMaximumLimit`.  We do not use `VeMinting` in this case, as in [bifrost kusama](https://github.com/bifrost-finance/bifrost/blob/d21077ad1c2376076702de5c4d4e1466024b6f12/runtime/bifrost-kusama/src/lib.rs#L1415)
 
 ### Changes in Node
 Partly adopted changes from [equilibrium's node service](https://github.com/eq-lab/equilibrium/blob/polkadot-v0.9.42/node/src/service/mod.rs). Changes are minor and mostly involve modifications in methods calls.

### DIFF
--- a/pallets/orml-tokens-management-extension/Cargo.toml
+++ b/pallets/orml-tokens-management-extension/Cargo.toml
@@ -11,29 +11,29 @@ serde = {version = "1.0.130", default-features = false, features = ["derive"], o
 sha2 = {version = "0.8.2", default-features = false}
 
 # Substrate dependencies
-frame-support = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.40", default-features = false}
-frame-system = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.40", default-features = false}
-sp-core = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.40", default-features = false}
-sp-runtime = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.40", default-features = false}
-sp-std = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.40", default-features = false}
+frame-support = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42", default-features = false}
+frame-system = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42", default-features = false}
+sp-core = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42", default-features = false}
+sp-runtime = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42", default-features = false}
+sp-std = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42", default-features = false}
 
-frame-benchmarking = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.40", default-features = false, optional = true }
+frame-benchmarking = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42", default-features = false, optional = true }
 
-orml-currencies = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.40", default-features = false }
-orml-tokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.40", default-features = false }
-orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.40", default-features = false }
+orml-currencies = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.42", default-features = false }
+orml-tokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.42", default-features = false }
+orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.42", default-features = false }
 
 
 
 [dev-dependencies]
 mocktopus = "0.8.0"
-frame-benchmarking = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.40" }
-sp-io = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.40", default-features = false}
+frame-benchmarking = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42" }
+sp-io = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42", default-features = false}
 
-pallet-balances = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.40"}
+pallet-balances = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42"}
 
 # Spacewalk libraries
-spacewalk-primitives = { git = "https://github.com/pendulum-chain/spacewalk", default-features = false, rev = "d05b0015d15ca39cc780889bcc095335e9862a36"}
+spacewalk-primitives = { git = "https://github.com/pendulum-chain/spacewalk", default-features = false, rev = "3c7d6c7f66f74ea1ea0d14cbe993c1e69856170b"}
 
 
 [features]
--- a/runtime/amplitude/src/lib.rs
+++ b/runtime/amplitude/src/lib.rs
@@ -366,7 +366,6 @@ impl Contains<RuntimeCall> for BaseFilter {
 			RuntimeCall::Farming(_) |
 			RuntimeCall::AssetRegistry(_) |
 			RuntimeCall::Proxy(_) |
-			RuntimeCall::VeMinting(_) |
 			RuntimeCall::RewardDistribution(_) => true,
 			// All pallets are allowed, but exhaustive match is defensive
 			// in the case of adding new pallets.
@@ -1267,35 +1266,9 @@ impl farming::Config for Runtime {
 	type Keeper = FarmingKeeperPalletId;
 	type RewardIssuer = FarmingRewardIssuerPalletId;
 	type FarmingBoost = FarmingBoostPalletId;
-	type VeMinting = VeMinting;
+	type VeMinting = ();
 	type BlockNumberToBalance = ConvertInto;
 	type WhitelistMaximumLimit = WhitelistMaximumLimit;
-}
-
-parameter_types! {
-	pub const VeMintingTokenType: CurrencyId = CurrencyId::Native;
-	pub VeMintingPalletId: PalletId = PalletId(*b"am/vemnt");
-	pub IncentivePalletId: PalletId = PalletId(*b"am/veict");
-	pub const Week: BlockNumber = 50400; // a week
-	pub const MaxBlock: BlockNumber = 10512000; // four years
-	pub const Multiplier: Balance = 10_u128.pow(12);
-	pub const VoteWeightMultiplier: Balance = 3;
-}
-
-impl bifrost_ve_minting::Config for Runtime {
-	type RuntimeEvent = RuntimeEvent;
-	type CurrencyId = CurrencyId;
-	type MultiCurrency = Currencies;
-	type ControlOrigin = EnsureRoot<AccountId>;
-	type TokenType = VeMintingTokenType;
-	type VeMintingPalletId = VeMintingPalletId;
-	type IncentivePalletId = IncentivePalletId;
-	type WeightInfo = ();
-	type BlockNumberToBalance = ConvertInto;
-	type Week = Week;
-	type MaxBlock = MaxBlock;
-	type Multiplier = Multiplier;
-	type VoteWeightMultiplier = VoteWeightMultiplier;
 }
 
 impl InstanceFilter<RuntimeCall> for ProxyType {
@@ -1423,7 +1396,6 @@ construct_runtime!(
 
 		// Asset Metadata
 		AssetRegistry: orml_asset_registry::{Pallet, Storage, Call, Event<T>, Config<T>} = 91,
-		VeMinting: bifrost_ve_minting::{Pallet, Call, Storage, Event<T>} = 92,
 
 		VestingManager: vesting_manager::{Pallet, Call, Event<T>} = 100
 	}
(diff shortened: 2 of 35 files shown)


runtime/amplitude/src/lib.rs
```diff
@@ -146,13 +147,67 @@ pub type SignedPayload = generic::SignedPayload<RuntimeCall, SignedExtra>;
...
+		if PolkadotXcm::on_chain_storage_version() == 0 {
+			log::info!("Upgrading pallet xcm's storage version to 1");
+			StorageVersion::new(1).put::<PolkadotXcm>();
+		}
+		if AssetRegistry::on_chain_storage_version() == 0 {
+			log::info!("Upgrading pallet asset registry's storage version to 2");
+			StorageVersion::new(2).put::<AssetRegistry>();
+		}
+
+		<Runtime as frame_system::Config>::DbWeight::get().reads_writes(8, 8)
```
ebma, line 195: Nit: since we have if statements guarding the `put` of the storage versions, we will not always have `8` writes. 
We could add an extra variable for counting the amount of storage writes by incrementing it in each if statement. 

This is not super critical probably but once this runtime upgrade caused all migrations to execute and update the storage versions we'll always only have `8` reads and `0` writes making the returned weight inaccurate.

ebma (commented):
I checked the migrations that you declared for Pendulum and Amplitude and they all make sense to me 👍 

I still have to check the other changes to the pallet configs tomorrow

ebma:
Do we also want to fix all issues on Foucoco or do we ignore it because it's just our test network? I can see that at least the transaction-payment version is also on `V1Ancient` on Foucoco, I didn't check all other pallets though.

runtime/amplitude/src/lib.rs
```diff
@@ -246,7 +322,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("amplitude"),
 	impl_name: create_runtime_str!("amplitude"),
 	authoring_version: 1,
-	spec_version: 13,
+	spec_version: 15,
```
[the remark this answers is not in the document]
gianfra-t: Actually it is a mistake. But I will leave it at 13 anyway so we can control the version and decouple it from this PR in particular.

pallets/orml-tokens-management-extension/Cargo.toml
```diff
@@ -11,29 +11,29 @@ serde = {version = "1.0.130", default-features = false, features = ["derive"], o
...
-sp-io = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.40", default-features = false}
+frame-benchmarking = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42" }
+sp-io = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42", default-features = false}
 
-pallet-balances = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.40"}
+pallet-balances = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42"}
 
 # Spacewalk libraries
-spacewalk-primitives = { git = "https://github.com/pendulum-chain/spacewalk", default-features = false, rev = "d05b0015d15ca39cc780889bcc095335e9862a36"}
+spacewalk-primitives = { git = "https://github.com/pendulum-chain/spacewalk", default-features = false, rev = "3c7d6c7f66f74ea1ea0d14cbe993c1e69856170b"}
```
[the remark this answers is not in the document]
gianfra-t: Yes! some spacewalk versions were pointing to different commits.

runtime/amplitude/src/lib.rs
gianfra-t: True, although this code should not live more than this upgrade, it is better practice for sure.

gianfra-t:
I think we should also fix the issues on Foucoco in case there are. But should we do this in this PR?

gianfra-t:
Okay so I looked at Foucoco and it seems like a good idea to bump here the versions that are behind, since they don't require much analysis.

Here is a brief breakdown:
- Pallet transaction-payment is `V1Ancient` but should be safe to bump for the same reason as in the other runtimes.
- Pallet vesting is at 0, but should be safe to bump since there is no vesting data!
- Pallet contracts is at 9. This one is different to the others but I think it is safe. If we look at what changed from 0.9.40 to 0.9.42 ([here](https://github.com/paritytech/substrate/compare/polkadot-v0.9.40...polkadot-v0.9.42#diff-70e9723e9db62816e35f6f885b6770a8449c75a6c2733e9fa7a245fe52c4656c) in `frame/contracts/src/lib.rs`), it seems that the storage type `DeletionQueue` was modified to a storage map. But right now there is no data if we query this store on-chain, and as long as there is nothing queued to be deleted when the migration happens, it should be good.
- Asset registry is also empty.

runtime/pendulum/src/xcm_config.rs
```diff
@@ -353,12 +357,90 @@ type Transactor = MultiCurrencyAdapter<
...
+		_: Self::Balance,
+	) -> Result<Option<Self::Balance>, DispatchError> {
+		todo!()
+	}
+
+	fn set_total_issuance(_: Self::AssetId, _: Self::Balance) {
+		todo!()
+	}
+}
+
```
ebma, line 437: Are we sure that we can leave `todo!()`?

I see [here](https://github.com/paritytech/cumulus/blob/polkadot-v0.9.42/primitives/utility/src/lib.rs#L124) that the trait bound only requires `fungibles::Mutate<AccountId> + fungibles::Balanced<AccountId>,`, is there a need to implement `Unbalanced` and `Inspect` as well?

runtime/pendulum/src/lib.rs
```diff
@@ -365,18 +408,42 @@ parameter_types! {
...
-	type DustRemoval = Treasury;
+	type DustRemoval = MoveDustToTreasury;
 	type ExistentialDeposit = ExistentialDeposit;
 	type AccountStore = System;
 	type WeightInfo = pallet_balances::weights::SubstrateWeight<Runtime>;
 	type MaxReserves = MaxReserves;
 	type ReserveIdentifier = ReserveIdentifier;
+	type FreezeIdentifier = ();
+	type MaxFreezes = ();
+	type MaxHolds = ConstU32<1>;
```
ebma, line 445: Is there a reason why we don't just set `type MaxFreezes = ConstU32<1>;` or similar as well?
I suppose you chose `type MaxHolds = ConstU32<1>;` arbitrarily or is there a reason for this exact value?

ebma (approved):
Looks good to me overall 👍  Thanks for the detailed description of this PR @gianfra-t, it really helped a lot when reviewing. 

@TorstenStueber please also have a look before we merge this

ebma:
@gianfra-t can you please resolve the conflicts of this PR? We need to prepare it for merging again

runtime/pendulum/src/xcm_config.rs
gianfra-t: I don't quite remember why this had to be implemented, even though the methods of the traits were not used.
In any case, with the new implementation of [fee per second](https://github.com/pendulum-chain/pendulum/pull/410), `TakeFirstAssetTrader` will not be used anymore, so the workaround is probably not needed anymore.

runtime/pendulum/src/lib.rs
gianfra-t: This is probably some of the configuration we need to agree on, but the `Get<u32>` default implementation for `()` is `0`, which would I think disallow the functionality. `MaxHolds` was also chosen arbitrarily.

gianfra-t:
The new [fee per second](https://github.com/pendulum-chain/pendulum/pull/410) implementation was merged into this branch, which removes the need for a temporary workaround used before.

runtime/pendulum/src/lib.rs
ebma: Ah okay, I didn't know that `Get<u32>` has an implementation for `()` but it makes sense. I think I'd prefer making this explicit with `ConstU32<0>` but it doesn't matter much.

prayagd:
@ebma @gianfra-t would this update break the ledger app?

gianfra-t:
I don't think so, because the extrinsics are not changing by the update (the transaction version should stay the same).

package-lock.json
ebma, line 1: Do we need this?

runtime/pendulum/src/lib.rs
```diff
@@ -1022,7 +1088,9 @@ impl dia_oracle::Config for Runtime {
 parameter_types! {
 	pub const FarmingKeeperPalletId: PalletId = PalletId(*b"pe/fmkpr");
 	pub const FarmingRewardIssuerPalletId: PalletId = PalletId(*b"pe/fmrir");
+	pub const FarmingBoostPalletId: PalletId = PalletId(*b"am/fmbst");
```
ebma, line 1091:
```suggestion
	pub const FarmingBoostPalletId: PalletId = PalletId(*b"pe/fmbst");
```

ebma (approved):
Looks good to me but should probably only be merged once https://github.com/pendulum-chain/spacewalk/pull/491 and https://github.com/pendulum-chain/pendulum/pull/410 are merged. The `rev` commits for the Spacewalk dependencies should then point to the merge commit of the PR on Spacewalk main.

package-lock.json
gianfra-t: Nope.

bogdanS98 (approved):
Looks good 👍🏼

gianfra-t:
Regarding merging this after https://github.com/pendulum-chain/pendulum/pull/410, do we plan to have both changes on the same runtime upgrade? I particularly have no strong opinion against doing both at the same time, and it could save some time. Also regarding [this](https://satoshipay.slack.com/archives/C01V1F56RMJ/p1710263548580969?thread_ts=1709826123.556089&cid=C01V1F56RMJ) slack discussion.

ebma:
Let's see if we can merge [this](https://github.com/pendulum-chain/spacewalk/pull/495) and then [this](https://github.com/pendulum-chain/spacewalk/pull/491) tomorrow so that we can link to a commit that is contained on the `main` branch of Spacewalk.

ebma (approved):

TorstenStueber (approved):
Let's go ahead with this. Thanks for all the testing work!

ebma:
Oh just one more thing. I noticed that the PR includes the changes of https://github.com/pendulum-chain/pendulum/pull/432 which we rolled back in order not to mess with the XCM configuration until https://github.com/pendulum-chain/tasks/issues/202 is fully ready and implemented. I propose we roll those changes back here as well. I assume this is quite simple and can be done with a simple revert of the specific commit like was done in https://github.com/pendulum-chain/pendulum/pull/431.

gianfra-t:
Oh, good that you remembered that @ebma, I completely forgot. Sadly I don't know if it will be that simple, mostly because there were some changes required for the old trader given the new version. So it will probably not look exactly like before.

gianfra-t:
@ebma I reverted the merge with fee per second, which re-introduces this workaround for the `TakeFirstAssetTrader`. When using `0.9.42`, simply using `Tokens` pallet instead of this `ConcreteAsset` struct defined in the workaround wont work, since it does not implement (anymore) the necessary traits. 
That I know, we only use the method `minimum_balance` from the inspect trait in `TakeFirstAssetTrader`. I removed the `todo()!` macro just to be safe so many implementations won't make sense.

Let me know what you think. I am testing with chopsticks right now.

ebma (approved):
Looks good to me, thanks for reverting the changes 👍 
Let's finally get this merged

PR #385, opened 2023-12-26
Add fee for incoming assets

gianfra-t:
Closes [#181](https://github.com/pendulum-chain/tasks/issues/181) 

We modify the `Trader` and `Barrier` implementations in order to be able to charge fees on the incoming asset sent. 

### Barrier

Barrier is modified from `AllowUnpaidExecutionFrom<Everything> to a set of [barriers](https://github.com/pendulum-chain/pendulum/pull/385/files#diff-f2b065c9a230c62a9e0320c5dc933302af3fe4c229d63a5432eb109e7a327e71R295).

The implementation for tuples allows the `xcm_executor` to iterate throughout all the barriers, until one returns Ok on it's implementation of `should_execute`.

The most important barrier to keep note of is `AllowTopLevelPaidExecutionFrom<Everything>`, which filters all XCM messages that do NOT contain an operation that deposits assets in the holding (like `WithdrawAsset` or `DepositReserve`) and that does NOT contain a `BuyExecution` with proper weight limit.

The other defined barriers are:
- `AllowKnownQueryResponses<PolkadotXcm>`: Allow messages that we are expecting (responses).
- `AllowSubscriptionsFrom<Everything>`: Version management and querying.
- `TakeWeightCredit`: Pay for the message execution with the environment credit.

### Trader
The implemented trader is [TakeFirstAssetTrader](https://github.com/pendulum-chain/pendulum/pull/385/files#diff-f2b065c9a230c62a9e0320c5dc933302af3fe4c229d63a5432eb109e7a327e71R327) defined [here](https://github.com/paritytech/cumulus/blob/d4bb2215bb28ee05159c4c7df1b3435177b5bf4e/primitives/utility/src/lib.rs#L120).

This trader will attempt to subtract the the fee in the first defined asset in the `BuyExecution` instruction, taking it from the holding registry. This implies that it must be deposited into it first by the sender.

The trader will first attempt to [convert](https://github.com/paritytech/cumulus/blob/d4bb2215bb28ee05159c4c7df1b3435177b5bf4e/primitives/utility/src/lib.rs#L166) the `MultiAsset` defined in the fee to a local asset id. It will then [call](https://github.com/paritytech/cumulus/blob/d4bb2215bb28ee05159c4c7df1b3435177b5bf4e/primitives/utility/src/lib.rs#L173) our [custom fee charger ](https://github.com/pendulum-chain/pendulum/pull/385/files#diff-f2b065c9a230c62a9e0320c5dc933302af3fe4c229d63a5432eb109e7a327e71R308) implementation to get the amount in the local asset id to be subtracted as fee payment.

To obtain this value, we make use of the base `amount` given by the [WeightToFee](hhttps://github.com/pendulum-chain/pendulum/pull/385/files#diff-f2b065c9a230c62a9e0320c5dc933302af3fe4c229d63a5432eb109e7a327e71R312) implementation, and adjust it by the [relative value](https://github.com/pendulum-chain/pendulum/pull/385/files#diff-f2b065c9a230c62a9e0320c5dc933302af3fe4c229d63a5432eb109e7a327e71R135) of that token vs the relay chain token which the [previous configuration](https://github.com/pendulum-chain/pendulum/pull/385/files#diff-f2b065c9a230c62a9e0320c5dc933302af3fe4c229d63a5432eb109e7a327e71L263) used 1:1 to charge fees in relay token. (this can be seen [here](https://github.com/paritytech/polkadot-sdk/blob/ac14d36514d9157121bffeb1ee4791c07d79c606/polkadot/xcm/xcm-builder/src/weight.rs#L227).
The fee collected will then be handled by the [custom fee handler](https://github.com/paritytech/cumulus/blob/d4bb2215bb28ee05159c4c7df1b3435177b5bf4e/primitives/utility/src/lib.rs#L269) which is defined here to be [transferred in it's entirety to the treasury](https://github.com/pendulum-chain/pendulum/pull/385/files#diff-f2b065c9a230c62a9e0320c5dc933302af3fe4c229d63a5432eb109e7a327e71R332)

### DIFF
--- a/runtime/amplitude/src/xcm_config.rs
+++ b/runtime/amplitude/src/xcm_config.rs
@@ -34,12 +34,6 @@ use cumulus_primitives_utility::{
 	ChargeWeightInFungibles, TakeFirstAssetTrader, XcmFeesTo32ByteAccount,
 };
 
-use super::{
-	AccountId, AmplitudeTreasuryAccount, Balance, Balances, Currencies, CurrencyId, ParachainInfo,
-	ParachainSystem, PolkadotXcm, Runtime, RuntimeCall, RuntimeEvent, RuntimeOrigin, Tokens,
-	Treasury, WeightToFee, XcmpQueue,
-};
-
 use crate::{
 	assets::{
 		native_locations::{native_location_external_pov, native_location_local_pov},
@@ -47,6 +41,13 @@ use crate::{
 	},
 	ConstU32,
 };
+
+use super::{
+	AccountId, AmplitudeTreasuryAccount, Balance, Balances, Currencies, CurrencyId, ParachainInfo,
+	ParachainSystem, PolkadotXcm, Runtime, RuntimeCall, RuntimeEvent, RuntimeOrigin, Tokens,
+	Treasury, WeightToFee, XcmpQueue,
+};
+
 
 parameter_types! {
 	pub const RelayLocation: MultiLocation = MultiLocation::parent();
@@ -82,7 +83,6 @@ impl Convert<CurrencyId, Option<MultiLocation>> for CurrencyIdConvert {
 			CurrencyId::XCM(index) => match index {
 				xcm_assets::RELAY_KSM => Some(MultiLocation::parent()),
 				xcm_assets::ASSETHUB_USDT => Some(asset_hub::USDT_location()),
-				9 => Some(MultiLocation::new(1, X2(Parachain(9999), PalletInstance(10)))),
 				_ => None,
 			},
 			CurrencyId::Native => Some(native_location_external_pov()),
@@ -101,8 +101,6 @@ impl Convert<MultiLocation, Option<CurrencyId>> for CurrencyIdConvert {
 			// Our native currency location with re-anchoring
 			// The XCM pallet will try to re-anchor the location before it reaches here
 			loc if loc == native_location_local_pov() => Some(CurrencyId::Native),
-			MultiLocation { parents: 1, interior: X2(Parachain(9999), PalletInstance(10)) } =>
-				Some(CurrencyId::XCM(9)),
 			_ => None,
 		}
 	}
@@ -117,6 +115,39 @@ impl Convert<MultiAsset, Option<CurrencyId>> for CurrencyIdConvert {
 		}
 	}
 }
+
+pub struct RelativeValue {
+	num: Balance,
+	denominator: Balance,
+}
+
+impl RelativeValue {
+	fn adjust_amount_by_relative_value(amount: Balance, relative_value: RelativeValue) -> Balance {
+		if relative_value.denominator == 0 {
+			// Or probably error
+			return amount
+		}
+		// Calculate the adjusted amount
+		let adjusted_amount = amount * relative_value.denominator / relative_value.num;
+		adjusted_amount
+	}
+}
+
+pub struct RelayRelativeValue;
+impl RelayRelativeValue {
+	fn get_relative_value(id: CurrencyId) -> Option<RelativeValue> {
+		match id {
+			CurrencyId::XCM(index) => match index {
+				xcm_assets::RELAY_KSM => Some(RelativeValue { num: 1, denominator: 1 }),
+				xcm_assets::ASSETHUB_USDT => Some(RelativeValue { num: 1, denominator: 4 }),
+				_ => None,
+			},
+			CurrencyId::Native => Some(RelativeValue { num: 1, denominator: 2 }),
+			_ => Some(RelativeValue { num: 1, denominator: 1 }),
+		}
+	}
+}
+
 
 /// Convert an incoming `MultiLocation` into a `CurrencyId` if possible.
 /// Here we need to know the canonical representation of all the tokens we handle in order to
@@ -248,7 +279,10 @@ impl ShouldExecute for DenyReserveTransferToRelayChain {
 		if matches!(origin, MultiLocation { parents: 1, interior: Here }) &&
 			instructions.iter().any(|inst| matches!(inst, ReserveAssetDeposited { .. }))
 		{
-			log::info!("{:?}", instructions);
+			log::trace!(
+				target: "xcm::barriers",
+				"Unexpected ReserveAssetDeposited from the relay chain",
+			);
 		}
 		// Permit everything else
 		Ok(())
@@ -274,21 +308,26 @@ pub type Barrier = (
 );
 
 pub struct ChargeWeightInFungiblesImplementation;
-
-//TODO some logic needs to be put in here. Most likely, we define weight_to_fee for all the assets
 impl ChargeWeightInFungibles<AccountId, Tokens> for ChargeWeightInFungiblesImplementation {
 	fn charge_weight_in_fungibles(
 		asset_id: CurrencyId,
 		weight: Weight,
 	) -> Result<Balance, XcmError> {
 		let amount = <WeightToFee as WeightToFeeTrait>::weight_to_fee(&weight);
-		log::info!("amount to be charger {:?} and asset {:?}", amount, asset_id);
-		Ok(amount)
+
+		if let Some(relative_value) = RelayRelativeValue::get_relative_value(asset_id) {
+			let adjusted_amount =
+				RelativeValue::adjust_amount_by_relative_value(amount, relative_value);
+			log::info!("amount to be charged: {:?} in asset: {:?}", adjusted_amount, asset_id);
+			return Ok(adjusted_amount)
+		} else {
+			log::info!("amount to be charged: {:?} in asset: {:?}", amount, asset_id);
+			return Ok(amount)
+		}
 	}
 }
 
 pub type Traders = (
-	UsingComponents<WeightToFee, RelayLocation, AccountId, Balances, ToAuthor<Runtime>>,
 	TakeFirstAssetTrader<
 		AccountId,
 		ChargeWeightInFungiblesImplementation,
--- a/runtime/amplitude/src/xcm_config.rs
+++ b/runtime/amplitude/src/xcm_config.rs
@@ -1,8 +1,22 @@
+use super::{
+	AccountId, AmplitudeTreasuryAccount, Balance, Balances, Currencies, CurrencyId, ParachainInfo,
+	ParachainSystem, PolkadotXcm, Runtime, RuntimeCall, RuntimeEvent, RuntimeOrigin, Tokens,
+	Treasury, WeightToFee, XcmpQueue,
+};
+
+use crate::{
+	assets::{
+		native_locations::{native_location_external_pov, native_location_local_pov},
+		xcm_assets,
+	},
+	ConstU32,
+};
 use core::marker::PhantomData;
 
 use frame_support::{
 	log, match_types, parameter_types,
 	traits::{ContainsPair, Everything, Nothing},
+	weights::{Weight, WeightToFee as WeightToFeeTrait},
 };
 use orml_traits::{
 	location::{RelativeReserveProvider, Reserve},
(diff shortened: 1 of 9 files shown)


runtime/pendulum/src/xcm_config.rs
```diff
@@ -253,6 +300,69 @@ impl ShouldExecute for DenyReserveTransferToRelayChain {
...
+	CurrencyIdConvert,
+	DepositToAlternative<PendulumTreasuryAccount, Currencies, CurrencyId, AccountId, Balance>,
+>;
+
+pub type Traders = (
+	TakeFirstAssetTrader<
+		AccountId,
+		ChargeWeightInFungiblesImplementation,
+		ConvertedConcreteId<CurrencyId, Balance, CurrencyIdConvert, JustTry>,
+		Tokens,
```
TorstenStueber, line 361: Here we should use Currencies so that we can also handle the case that someone uses our native asset for fee payments.

runtime/pendulum/src/xcm_config.rs
```diff
@@ -92,6 +100,45 @@ impl Convert<CurrencyId, Option<MultiLocation>> for CurrencyIdConvert {
...
+				xcm_assets::ASSETHUB_USDT => Some(RelativeValue { num: 1, denominator: 4 }),
+				xcm_assets::ASSETHUB_USDC => Some(RelativeValue { num: 1, denominator: 4 }),
+				xcm_assets::EQUILIBRIUM_EQD => Some(RelativeValue { num: 1, denominator: 10 }),
+				xcm_assets::MOONBEAM_BRZ => Some(RelativeValue { num: 1, denominator: 10 }),
+				xcm_assets::POLKADEX_PDEX => Some(RelativeValue { num: 1, denominator: 2 }),
+				xcm_assets::MOONBEAM_GLMR => Some(RelativeValue { num: 1, denominator: 10 }),
+				_ => None,
+			},
+
+			CurrencyId::Native => Some(RelativeValue { num: 1, denominator: 2 }),
```
TorstenStueber, line 135: Wouldn't it make more sense to use denominator 1 here?

TorstenStueber (commented):
Great PR, I have just a few minor comments.

Some other comments:
- You mention the `TakeWeightCredit` barrier in your description but don't use it
-

runtime/pendulum/src/xcm_config.rs
gianfra-t: Well in this example you point out, our native would be "worth" half of what the relay chain token is. All the relative valuations are examples. If we decide to go with this solution then we should plug in the adequate fraction.

runtime/pendulum/src/xcm_config.rs
gianfra-t: Totally agree, but there is a problem, `Currencies` does not implement the necessary traits like `fungibles::Transfer`..
In any case, `Tokens` actually makes use of the type `ExistentialDeposits` which will always return `NANOUNIT`, so it can also handle the native asset. 
You can actually see that in [this modified test](https://github.com/pendulum-chain/pendulum/pull/385/files#diff-3429c5361e6214779992604d4212c994d6c327c55b74c9bcb280681ad5ddf93dR617) the treasury is getting the fee in Native.

runtime/amplitude/src/xcm_config.rs
```diff
@@ -244,7 +285,53 @@ impl ShouldExecute for DenyReserveTransferToRelayChain {
...
+		MultiLocation { parents: 1, interior: Here } |
+		MultiLocation { parents: 1, interior: X1(Plurality { .. }) }
+	};
+}
+
+pub type Barrier = (
+	TakeWeightCredit,
+	AllowTopLevelPaidExecutionFrom<Everything>,
+	// Parent and its plurality get free execution
+	AllowUnpaidExecutionFrom<ParentOrParentsPlurality>,
```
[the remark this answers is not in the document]
gianfra-t: To be honest, I do not know what exact message the relay chain can send to us. But `ParentOrParentsPlurality` will not allow an origin multi-location with an address in the interior, so I don't think anyone can spam us from the relay. Only the relay itself or `Plurality` which as far as I understand is the council for example, but I could be wrong on that last bit.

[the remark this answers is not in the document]
TorstenStueber: I don't think that so. If someone sends a message from any other chain, the origin multilocation will still point to the relay chain or parachain, the only difference will be the first instruction that will be `DescendOrigin`, see here: https://github.com/pendulum-chain/pendulum/pull/365#issuecomment-1863492020

runtime/pendulum/src/xcm_config.rs
TorstenStueber: Oh, seems that you are right. I am almost shocked that `Currency` does not implement `fungibles`, that would be actually an important point. So how do we want to deal with a situation where someone wants to use our native asset to pay fees?

runtime/pendulum/src/xcm_config.rs
TorstenStueber: As I point out in [this comment](https://github.com/pendulum-chain/pendulum/pull/385#pullrequestreview-1798251341), I think that the base fee is in terms of our native asset, not in the relay chain native asset.

gianfra-t: Oh okay, I see. Then we should have to modify our [base fee](https://github.com/pendulum-chain/pendulum/pull/385/files#diff-f2b065c9a230c62a9e0320c5dc933302af3fe4c229d63a5432eb109e7a327e71R312) to be in terms of the native token. In that case yes, the relation of course is 1:1.
Right now we are actually using the value given by `WeightToFee` as the fee in relay token, this was defined [here](https://github.com/pendulum-chain/pendulum/pull/385/files#diff-f2b065c9a230c62a9e0320c5dc933302af3fe4c229d63a5432eb109e7a327e71L263).

runtime/pendulum/src/xcm_config.rs
gianfra-t: As long we implement like [here](https://github.com/pendulum-chain/pendulum/blob/main/runtime/pendulum/src/lib.rs#L686) the existential deposit function, the tokens module will be able to give the minimum_balance and we can charge in native, as in the test.
I agree it is a bit confusing, but I cannot think of a good alternative right now.

runtime/pendulum/src/xcm_config.rs
TorstenStueber: I think was wrong to begin with. The gold standard (I think) is how pallet transaction payment does it and how it assigns a fee in our native asset to a particular weight.

Pallet-transaction-payment is more complicated of course and also uses an adaptive fee multiplier, which at low activity is well below 1. But assuming a multiplier of 1 and neglecting tips and length fees, then `WeightToFee` is exactly the native asset fee assigned to an extrinsic of a specific weight. More here: https://www.notion.so/satoshipay/22-05-26-Transaction-Payment-dd2304d17baa4b6d9ae0b76873069870?pvs=4

runtime/amplitude/src/xcm_config.rs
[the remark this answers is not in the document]
gianfra-t: Oh I thought that `DescendOrigin` was added as well.
In that case I guess we could use a custom barrier that filters for `DescendOrigin` instructions, yet allows everything else from the relay chain.
I am surprised that `AllowUnpaidExecutionFrom<ParentOrParentsPlurality>` is widely used in other chains. I'm going to try and dig a bit why would that be the case.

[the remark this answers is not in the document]
TorstenStueber: Maybe the explanation is that the relay chain has some mechanism to disallow that users send arbitrary XCM messages to parachains?

runtime/pendulum/src/xcm_config.rs
TorstenStueber: Seems like you are right in that the only components `TakeFirstAssetTrader` uses from this fourth generic (internally called `ConcreteAssets`) is the `AssetId`, the `Balance` and the `minimum_balance`. For the former two `Tokens` already gives the right value.

For the `minimum_balance` would just report its own value `NANOUNIT` and not the actual minimum balance we set up in pallet-balances, which is [a milliunit instead](https://github.com/pendulum-chain/pendulum/blob/fbb86dc4d27984a44abcb9ef72e89768183f238f/runtime/common/src/lib.rs#L45).

This way our `TakeFirstAssetTrader` implementation might not work correctly as it assumes a wrong minumum balance. Therefore I think that we should rewrite [our `ExistentialDeposits` function](https://github.com/pendulum-chain/pendulum/blob/fbb86dc4d27984a44abcb9ef72e89768183f238f/runtime/pendulum/src/lib.rs#L686) and explicitly return the same `EXISTENTIAL_DEPOSIT` we use for pallet-balances whenever the argument `_currency_id` is the native asset.

runtime/amplitude/src/xcm_config.rs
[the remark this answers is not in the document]
TorstenStueber: Ah, got it: the `send` extrinsic of `pallet-xcm` requires [a specific origin](https://github.com/paritytech/polkadot-sdk/blob/ae14e6da6a3b2a729365f9581211d07033719c1a/polkadot/xcm/pallet-xcm/src/lib.rs#L329C29-L329C42) that can be freely configured.

The relay chain runtime configured this to be only ["the root, the general admin, the fellowship admin and the staking admin"](https://github.com/polkadot-fellows/runtimes/blob/30e0dbfdcb78722ed61325c0ebf1efdcdb6033ba/relay/polkadot/src/xcm_config.rs#L417-L419).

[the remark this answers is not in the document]
TorstenStueber: So then we can leave this here.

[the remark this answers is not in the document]
gianfra-t: Oh, nice observation. Thanks!

gianfra-t:
I modified the relative valuations (roughly accurate), yet could not find the price for AMPE. 
As for  `TakeWeightCredit` I'm sorry I missed the comment. This barrier is set as one of the multiple barriers we use. It's purpose as far as I understand is to allow local messages to be sent. 

The barrier is simple. It just attempts to consume some credit that is passed when calling `execute_xcm_in_credit`. When  `execute_xcm` is called, it just actually calls `execute_xcm_in_credit` with 0 credit [see here](https://github.com/paritytech/polkadot-sdk/blob/8bf5a1c0b31a3bfa4751c50a374e3ce3d4cfda1d/polkadot/xcm/src/v3/traits.rs#L356).
I believe the intention of this barrier is so that we can call  `execute_xcm_in_credit` with a message coming from inside the runtime, like [here](https://github.com/pendulum-chain/bifrost/blob/71d2e5077de0c3b7e785dc2ea39b622300b7307b/pallets/slp/src/agents/common.rs#L265), where we can define the credit we want.

TorstenStueber:
Oh, of course, I missed that you actually do you `TakeWeightCredit`. Sure, it is meant for internal use, e.g., when initiating a reserveAssetTransfer from our chain, then the xcm message that will be executed locally will take the weight credit from the transaction fee, so that it totally fine.

TorstenStueber (approved):
Looks all good now.

runtime/pendulum/src/xcm_config.rs
```diff
@@ -92,6 +100,45 @@ impl Convert<CurrencyId, Option<MultiLocation>> for CurrencyIdConvert {
...
+}
+
+impl RelativeValue {
+	fn adjust_amount_by_relative_value(amount: Balance, relative_value: RelativeValue) -> Balance {
+		if relative_value.denominator == 0 {
+			// Or probably error
+			return amount
+		}
+		// Calculate the adjusted amount
+		let adjusted_amount = amount * relative_value.denominator / relative_value.num;
```
ebma, line 115: Shouldn't we rather be defensive here and use functions like [saturating_mul](https://docs.rs/sp-runtime/latest/sp_runtime/traits/trait.Saturating.html#) and [checked_div](https://docs.rs/sp-runtime/latest/sp_runtime/traits/trait.CheckedDiv.html) to prevent the runtime from panicking?

Also, I think it's a bit counter-intuitive to multiply with a denominator and divide by a numerator although it works. Don't have a suggestion for better naming though.

gianfra-t: Yes I overlook the saturating operators, I think it is a good idea to add it. I can append the changes to [this PR](https://github.com/pendulum-chain/pendulum/pull/386).

And about the naming, it can be confusing on the operation, but on the definition of the rational number for the value ratio makes sense to call them that way.

gianfra-t: I added the changes [here](https://github.com/pendulum-chain/pendulum/pull/386/files#diff-3287a4ee060b1070a8e6559a094b4890a596a1b281effef54cafe59da20edf49R79). Also moved the type `RelativeValue` itself to common.

Issue #327, opened 2023-10-21
Chain not compiling 


siddharthteli12:
- Getting error **unresolved import `primitives::stellar::TransactionSetType`**,  due to `stellar-relay` in amplitude runtime. 

  https://github.com/pendulum-chain/pendulum/blob/35807d58564ab06b1a3647b7335dd81e5c43d88d/runtime/amplitude/Cargo.toml#L34

- `TransactionSetType` is define [here](https://github.com/pendulum-chain/substrate-stellar-sdk/blob/740389c7ec47a8725c32d8a68061ffb1bdfd013b/src/xdr/impls/transaction_set_type.rs#L9-L12)
- Pallet `stellar-relay` import needs to be fixed [here](https://github.com/pendulum-chain/spacewalk/blob/c2e524515fddd16c9618ff0c7c29ba2aba3f994b/pallets/stellar-relay/src/lib.rs#L40C30-L40C48)

TorstenStueber:
@ebma can you confirm?

siddharthteli12:
@TorstenStueber , I am not getting the above error right now. But when I am building chain with `--all-features` flag the chain is not compiling. Specifically, The error is originating from `try-runtime` feature flag. 

Apart from that I observed all previous merge's are passing GitHub actions because we are only running `cargo check` we should at least run `cargo check --all-features`. 

Will also recommend running `cargo clippy`  later after everything is sorted.

ebma:
Yes, I can confirm that the runtimes throw errors when trying to compile them with the `try-runtime` feature. I renamed the title of this issue because that's what we need to fix. 

I'm not so sure if we should add the `cargo check --all-features` to our CI workflows though. I briefly checked the workflows of other parachain teams and it doesn't seem like they do. We might have features that are mutually exclusive and it wouldn't work to compile with all features enabled. But we can see if that is the case once the `try-runtime` compilation is fixed.

TorstenStueber:
Yes, that's what I think. I would assume that `cargo check --all-features` is not a meaningful thing to do.

vadaynujra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/pendulum/327?planning-poker) with Zenhub @b-yap @bogdanS98 @ebma @gianfra-t @TorstenStueber

gianfra-t:
[...]

PR #386, opened 2023-12-29
fix for try-runtime features

gianfra-t:
This PR adds fixes so that the runtimes can be compiled with the `try-runtime` feature. Also should solve #327.
Related to [tasks/#64](https://github.com/pendulum-chain/tasks/issues/64) and [the upgrade to 0.9.42](https://github.com/pendulum-chain/pendulum/pull/382). I believe it would be a good idea to first have the capability of testing with `try-runtime` the upgrade.

This PR also has a small fix regarding [this comment](https://github.com/pendulum-chain/pendulum/pull/385#discussion_r1447475914)


ebma:
Would you say this is related to #327 then?

gianfra-t:
Yes! I wasn't aware about that issue. Adding it to the description.

TorstenStueber:
@gianfra-t since the [ticket](https://github.com/pendulum-chain/spacewalk/pull/472) you mention in your ticket above has been closed, can we open this PR here for review or is it still a draft?

gianfra-t:
I'm right now merging with main and checking that everything still works.

runtime/development/src/lib.rs
```diff
@@ -583,23 +583,20 @@ impl_runtime_apis! {
...
-			block: Block,
-			state_root_check: bool,
-			signature_check: bool,
-			select: frame_try_runtime::TryStateSelect
-		) -> Weight {
-			// NOTE: intentional unwrap: we don't want to propagate the error backwards, and want to
-			// have a backtrace here.
-			Executive::try_execute_block(block, state_root_check, signature_check, select).expect("execute-block failed")
-		}
+            block: Block,
```
TorstenStueber, line 591: Can you please run rust fmt?

runtime/common/src/lib.rs
```diff
@@ -70,6 +70,23 @@ pub mod opaque {
...
+
+pub struct RelativeValue<Amount> {
+	pub num: Amount,
+	pub denominator: Amount,
+}
+
+impl<Amount: CheckedDiv<Output = Amount> + Saturating + Clone> RelativeValue<Amount> {
+	pub fn adjust_amount_by_relative_value(amount: Amount, relative_value: RelativeValue<Amount>) -> Amount {
+		// Calculate the adjusted amount
+		if let Some(adjusted_amount) = amount.clone().saturating_mul(relative_value.denominator).checked_div(&relative_value.num) {
```
TorstenStueber, line 82: As @ebma already pointed out, it is not intuitive that this is multiplied by the denominator and divided by the numerator.

This is just because I would expect it more that a multiplication with the fraction `RelativeValue` happens here, but instead it is a division – which is also fine. But maybe we can make it more explicit by calling this function `divide_by_relative_value` instead of just `adjust_...`?

TorstenStueber (commented):
There is also a Spacewalk dependency in `pallets/orml-tokens-management-extension/Cargo.toml` and in `runtime/integration-tests/Cargo.toml`, which should probably also be updated (?)

runtime/common/src/lib.rs
gianfra-t: sure!

gianfra-t:
> There is also a Spacewalk dependency in pallets/orml-tokens-management-extension/Cargo.toml and in runtime/integration-tests/Cargo.toml, which should probably also be updated (?)

Thanks for finding this. It mus have been a reference to an even older commit, not sure how it was compiling!

TorstenStueber (approved):
Thanks, looks all good to me now.

TorstenStueber:
However, would be good to also get @ebma's opinion here as he commented [here](https://github.com/pendulum-chain/pendulum/pull/385#discussion_r1447475914) on the changes with the `RelativeValue`.

runtime/pendulum/Cargo.toml
```diff
@@ -224,6 +225,51 @@ runtime-benchmarks = [
...
+    "pallet-balances/try-runtime",
+    "pallet-bounties/try-runtime",
+    "pallet-child-bounties/try-runtime",
+    "pallet-collective/try-runtime",
+    "pallet-contracts/try-runtime",
+    "pallet-democracy/try-runtime",
+    "pallet-identity/try-runtime",
+	  "pallet-multisig/try-runtime",
+    "pallet-preimage/try-runtime",
+	  "pallet-proxy/try-runtime",
```
ebma, line 244: The indentation here and in the other Cargo.toml file is a bit off

ebma (approved):
LGTM

runtime/pendulum/Cargo.toml
gianfra-t: Strange that cargo format doesn't fix this automatically.

gianfra-t: Also strange that locally I cannot see the wrong indentation!

ebma: I think Cargo.toml only checks the .rs files and no .toml files unfortunately 🤔

ebma (approved):

PR #396, opened 2024-01-11
Implement pallet that extends our current treasury pallet with buyout feature

### LINKED ISSUE #395: Implement pallet that extends our current treasury pallet with buyout feature
Implement pallet that extends our current treasury pallet with buyout feature based on Equilibrium's treasury [pallet](https://github.com/eq-lab/equilibrium/tree/9836ce327f1a24ca4c2b81cd4787e4eb90c707ad/pallets/eq-treasury) and the research done https://github.com/pendulum-chain/tasks/issues/113 using approach described [here](https://www.notion.so/satoshipay/23-08-16-Equilibrium-Treasury-0d27bcaa0a0d4d99ae62a353ff553c67?pvs=4).

Asset prices should be fetched by using Spacewalk's [oracle](https://github.com/pendulum-chain/spacewalk/tree/main/pallets/oracle) pallet

bogdanS98:
Closes #395. 

### Overview

This pallet allows users which hold other assets than our native asset to exchange them against the treasury in return for our native. This is limited by the assets supported, minimum and maximum buyout amounts and buyout period (after which the total buyout amount for a user is reset).

Buyout limit can only be changed by root. If set to None, then there is no amount limit for buyouts.

This pallet was only added for Foucoco runtime in this PR.

### Assets supported

Amplitude

- USDT
- KSM

Pendulum

- USDC
- USDT
- DOT
- GLMR

### Notes: 

1. We still have to properly define the values of the following parameters for each runtime:
  - `SellFee`
  - `MinAmountToBuyout`
  - `BuyoutPeriod`
  
2. This pallet has yet to be tested in Polkadot.js


pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+				Some(acc) =>
+					if acc == treasury_account_id {
+						Error::<T>::InsufficientTreasuryBalance.into()
+					} else if acc == who {
+						Error::<T>::InsufficientAccountBalance.into()
+					} else {
+						error
+					},
+				_ => error,
+			})?;
```
ebma, line 307: Maybe it's best if we don't extract the logic for 'exchanging' the tokens into an extra function. It's done like this by equilibrium as they implement this as a primitive on their balances pallet. But to be honest, I'm not a huge fan of that function overall, and I find it to be more confusing than helping with readability.

So I think it's best if instead you put the logic that is now hidden in the `exchange()` function to here. Then we can remove the `ExchangeFailure` error type and we also don't need to deposit the `Exchange` event IMO as calling the `transfer()` function on the `Currencies` trait will emit similar events anyways.

ebma (commented):
I left some comments. 

Overall I think we should never use '+,-,*,/' operators directly and rather use defensive operations like 'saturating_..' or 'checked_..'.
And I think we can remove the tight coupling to the oracle pallet. 

Besides that we also need the logic for the signed extension but I know you are aware of this and it's still a WIP PR after all.

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -0,0 +1,354 @@
...
+				Rounding::NearestPrefDown,
+			)
+		}
+
+        // Use NoPrice error here maybe
+		fn fetch_prices(
+			assets: (&CurrencyIdOf<T>, &CurrencyIdOf<T>),
+		) -> Result<(FixedU128, FixedU128), DispatchError> {
+			let basic_asset_price: FixedU128 =
+				T::PriceGetter::get_price::<FixedU128>(*assets.0)?.into();
```
gianfra-t, line 338: Here it will [need](https://github.com/pendulum-chain/pendulum/pull/396/files#diff-b2aa68998afa0af43db27d3ee43b0298c6d614ee556a1b60969a00a9fe514696R247) to return the price of the native asset but with the fee, is this something we will add to directly to the implementation of the `PriceGetter`? 

Otherwise we could use the `SellerFee` getter to calculate (+one and multiply) the value with fee here. I see that equilibrium does it [here](https://github.com/eq-lab/equilibrium/blob/ddab6da9fdbca80c7988b21c3eff96712d576f64/pallets/eq-treasury/src/lib.rs#L299)

bogdanS98: True, re-added that now since I lost that part while refactoring

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -0,0 +1,525 @@
...
+		buyout_amount: BalanceOf<T>,
+	) -> DispatchResult {
+		if let Some(buyout_limit) = BuyoutLimit::<T>::get() {
+			let buyout_period = T::BuyoutPeriod::get();
+			// Get current block number
+			let now = <frame_system::Pallet<T>>::block_number().saturated_into::<u32>();
+			let current_period = now
+				.checked_div(buyout_period)
+				.and_then(|n| Some(n.saturating_mul(buyout_period)))
+				.unwrap_or_default();
```
ebma, line 192: What actually is the expected outcome here? Rounding? We are dividing and multiplying with the same value after all.

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -0,0 +1,525 @@
...
+			if !buyouts.is_zero() && last_buyout < current_period {
+				buyouts = Default::default();
+				Buyouts::<T>::insert(account_id, (buyouts, now));
+			};
+
+			ensure!(
+				buyouts
+					.saturated_into::<u128>()
+					.saturating_add(buyout_amount.saturated_into::<u128>()) <=
+					buyout_limit.saturated_into::<u128>(),
```
ebma, line 204: Do we have to do the `saturated_into()` calls? Is there a way to safely add `Balance` types without casting here?

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -0,0 +1,525 @@
...
+		/// Buyout period in blocks
+		#[pallet::constant]
+		type BuyoutPeriod: Get<u32>;
+
+		/// Fee from the native asset buyouts
+		#[pallet::constant]
+		type SellFee: Get<Permill>;
+
+		/// Type that allows for checking if currency type is ownable by users
+		type AllowedCurrencyIdVerifier: AllowedCurrencyIdVerifier<CurrencyIdOf<Self>>;
```
ebma, line 70: The name is a bit clunky. Is `AllowedCurrencyChecker` better? Not sure

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -0,0 +1,525 @@
...
+				Error::<T>::BuyoutLimitExceeded
+			);
+		}
+
+		Ok(())
+	}
+
+	/// Ensures that asset is allowed for buyout
+	/// The concrete implementation of AllowedCurrencyIdVerifier trait must be provided by the runtime
+	fn ensure_allowed_asset_for_buyout(asset: &CurrencyIdOf<T>) -> DispatchResult {
```
ebma, line 214:
```suggestion
	fn ensure_asset_allowed_for_buyout(asset: &CurrencyIdOf<T>) -> DispatchResult {
```

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -0,0 +1,525 @@
...
+		);
+
+		Ok(())
+	}
+
+	/// Updates buyouts storage for the account
+	fn update_buyouts(account_id: &AccountIdOf<T>, buyout_amount: BalanceOf<T>) {
+		if BuyoutLimit::<T>::get().is_some() {
+			Buyouts::<T>::mutate(account_id, |(prev_buyouts, last)| {
+				*prev_buyouts = *prev_buyouts + buyout_amount;
```
ebma, line 227: Nit: use saturating add

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -0,0 +1,525 @@
...
+		buyout_amount: BalanceOf<T>,
+	) -> Result<BalanceOf<T>, DispatchError> {
+		let basic_asset = <T as orml_currencies::Config>::GetNativeCurrencyId::get();
+		ensure!(asset != basic_asset, Error::<T>::WrongAssetToBuyout);
+
+		let (basic_asset_price, exchange_asset_price) = Self::fetch_prices((&basic_asset, &asset))?;
+
+		// Add fee to the basic asset price
+		let basic_asset_price_with_fee =
+			basic_asset_price * (FixedU128::from(T::SellFee::get()) + FixedU128::one());
```
ebma, line 245: Nit: also here safe calculation operations

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -0,0 +1,525 @@
...
+		let basic_asset = <T as orml_currencies::Config>::GetNativeCurrencyId::get();
+		let (buyout_amount, exchange_amount) = Self::split_to_buyout_and_exchange(asset, amount)?;
+
+		Self::ensure_buyout_limit_not_exceeded(&who, buyout_amount)?;
+		let treasury_account_id = T::TreasuryAccount::get();
+
+		// Start exchanging
+		// Check for exchanging zero values and same accounts
+		if exchange_amount.is_zero() && buyout_amount.is_zero() || who == treasury_account_id {
+			return Ok(())
```
ebma, line 320: Should we return an error here?

ebma: Or maybe two different errors even

runtime/foucoco/src/lib.rs
```diff
@@ -973,6 +995,80 @@ impl orml_tokens_management_extension::Config for Runtime {
...
 }
 
+pub struct AllowedCurrencyIdVerifierImpl;
+impl treasury_buyout_extension::AllowedCurrencyIdVerifier<CurrencyId>
+	for AllowedCurrencyIdVerifierImpl
+{
+	fn is_allowed_currency_id(currency_id: &CurrencyId) -> bool {
+		matches!(
+			currency_id,
+			CurrencyId::XCM(0) | CurrencyId::XCM(1) | CurrencyId::XCM(2) | CurrencyId::XCM(6)
```
ebma, line 1005: Should we just allow all assets? Doesn't matter for the benchmarks but is less error prone

runtime/foucoco/src/lib.rs
```diff
@@ -973,6 +995,80 @@ impl orml_tokens_management_extension::Config for Runtime {
...
+		)
+	}
+}
+pub struct OracleWrapper(Oracle);
```
ebma, line 1009: Maybe better:
```suggestion
pub struct OraclePriceGetter(Oracle);
```

ebma (changes requested):
Looks good overall, nice job @bogdanS98 👍 

We might have to check what exactly we need to do for the SignedExtensions again, I think they need more configuration than that (in the runtime).

pallets/treasury-buyout-extension/src/lib.rs
gianfra-t: I think it is because of rounding. First division if it has a remainder it will get truncated, multiplying by the number again will get you a smaller number. If one period has passed (lets say exactly), then it will not get truncated, which is then used int he condition bellow.
To me, it is the same as doing `now - buyout_period > last_buyout`, is it @bogdanS98 ?

runtime/foucoco/src/lib.rs
```diff
@@ -973,6 +995,80 @@ impl orml_tokens_management_extension::Config for Runtime {
...
+		FixedNumber: FixedPointNumber + One + Zero + Debug + TryFrom<FixedU128>,
+	{
+		// Forcefully set chain status to running when benchmarking so that the oracle doesn't fail
+		Security::set_status(StatusCode::Running);
+
+		let key = OracleKey::ExchangeRate(currency_id);
+		let rate = FixedU128::checked_from_rational(100, 1).expect("This is a valid ratio");
+		// Needed for feeding values to the oracle but is never used in the oracle
+		let account = AccountId::from([0u8; 32]);
+		Oracle::feed_values(account, vec![(key.clone(), rate)]);
```
gianfra-t, line 1038: Every time `get_price` gets called we are also feeding the values, so in the benchmarks we would be measuring this operation which won't happen in the non-benchmark case. 

I don't know how much the detail of the benchmark really matters, but perhaps we could have a function `feed_price` that is only used with `runtime-benchmarks` instead of two `get_prices`, and in the benchmarks code we can just call `feed_price` before the call to the extrinsic. WDYT?

gianfra-t (commented):
Everything looks very nice! 
Just left some comments but are more details. Last thing that I think we need to add for this to work is what @ebma said, and declaring the `SignedExtension` check in the runtime.

I think [this](https://github.com/pendulum-chain/pendulum/blob/4ea070519ce59df4bce3d316c8ef2e6533da41b6/runtime/amplitude/src/lib.rs#L130) is the place we have to add it, but not sure if this is the only place.

runtime/foucoco/src/lib.rs
ebma: Scratch that, I forgot that this setting is actually important for our runtime. When I wrote the comment I assumed that it's mostly used in the benchmark.

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -0,0 +1,525 @@
...
+		/// Currency id of the relay chain
+		#[cfg(feature = "runtime-benchmarks")]
+		type RelayChainCurrencyId: Get<CurrencyIdOf<Self>>;
+	}
+
+	#[pallet::pallet]
+	pub struct Pallet<T>(PhantomData<T>);
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
```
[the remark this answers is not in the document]
bogdanS98: I agree with this, it's more readable to have the storage items, events and errors before the extrinsics which use them 👍🏼

pallets/treasury-buyout-extension/src/lib.rs
bogdanS98: Yes, it is rounding. We use this to get the block when the current period started. For example if current block is `150000` and the buyout period is `7200` blocks then `current_period = (150000 / 7200) * 7200` which after the rounded division will become `current_period = 20 * 7200` therefore `current_period = 144000 `. Maybe I should refactor `current_period` and call it `current_period_start_block` to be clearer.

pallets/treasury-buyout-extension/src/lib.rs
bogdanS98: Yes, indeed it's a bit clunky and your suggestion sounds good. I wanted this name to not be easily confused with `CurrencyIdChecker` from the `orml-tokens-management-extension` pallet

pallets/treasury-buyout-extension/src/lib.rs
bogdanS98: Yes, we should definitely return 2 different errors 👍🏼

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -0,0 +1,525 @@
...
+		/// Allows caller to buyout a given amount of native token.
+		/// Caller can specify either buyout amount of native token that he wants or exchange amount of an allowed asset.
```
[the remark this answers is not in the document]
bogdanS98: Done 👍🏼

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -0,0 +1,525 @@
...
+			who: AccountIdOf<T>,
+			buyout_amount: BalanceOf<T>,
+			asset: CurrencyIdOf<T>,
+			exchange_amount: BalanceOf<T>,
+		},
+		/// Buyout limit updated event
+		BuyoutLimitUpdated { limit: Option<BalanceOf<T>> },
+	}
+
+	/// Stores limit amount user could by for a period.
```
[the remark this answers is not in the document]
bogdanS98: Done 👍🏼

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -0,0 +1,525 @@
...
+		/// Used for currency-related operations
+		type Currency: MultiCurrency<AccountIdOf<Self>, CurrencyId = CurrencyIdOf<Self>>;
+
+		/// Used for getting the treasury account
+		#[pallet::constant]
+		type TreasuryAccount: Get<AccountIdOf<Self>>;
+
+		/// Buyout period in blocks
+		#[pallet::constant]
+		type BuyoutPeriod: Get<u32>;
```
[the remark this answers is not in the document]
bogdanS98: Done 👍🏼

pallets/treasury-buyout-extension/src/lib.rs
bogdanS98: You're right, we don't really need the casting. Updated the code 👍🏼

pallets/treasury-buyout-extension/src/lib.rs
bogdanS98: @gianfra-t It is not actually the same as `now - buyout_period > last_buyout` since we're calculating which is the block number at which the last buyout period started and checking if the last buyout was executed at some block before that so that we can reset the buyout amount limit for that caller. If we were doing `now - buyout_period > last_buyout` then we would always assume that `now` is the end of a buyout period which is not always true.

gianfra-t: No I was not suggesting to change! Like you say it is similar but not exactly the same, and the solution implemented it makes more sense if we think in terms of periods.

ebma:
Is this ready for re-review now @bogdanS98?

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -0,0 +1,525 @@
...
+		who: AccountIdOf<T>,
+		asset: CurrencyIdOf<T>,
+		amount: Amount<BalanceOf<T>>,
+	) -> DispatchResult {
+		Self::ensure_allowed_asset_for_buyout(&asset)?;
+
+		let basic_asset = <T as orml_currencies::Config>::GetNativeCurrencyId::get();
+		let (buyout_amount, exchange_amount) = Self::split_to_buyout_and_exchange(asset, amount)?;
+
+		Self::ensure_buyout_limit_not_exceeded(&who, buyout_amount)?;
```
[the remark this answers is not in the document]
bogdanS98: Good point, but I haven't found yet what could be the implications of removing the checks from the extrinsics. We could remove them but we need to make sure that there's no way to bypass the `SignedExtension` and call the extrinsic.

runtime/foucoco/src/lib.rs
bogdanS98: This is a good point 👍🏼 However, I think that if we do it this way, benchmarks will become tightly coupled to the Oracle pallet and I'm not so sure how big the differences will be benchmarking wise. Is it worth doing this? Maybe @ebma has a different opinion

pallets/treasury-buyout-extension/src/lib.rs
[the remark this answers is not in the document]
gianfra-t: Yes, I was assuming there was no way of doing this but that is a dangerous assumption.

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -79,18 +80,64 @@ pub mod pallet {
...
+	#[pallet::storage]
+	pub type Buyouts<T: Config> =
+		StorageMap<_, Blake2_128Concat, AccountIdOf<T>, (BalanceOf<T>, u32), ValueQuery>;
 
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
 		/// Allows caller to buyout a given amount of native token.
-		/// Caller can specify either buyout amount of native token that he wants or exchange amount of an allowed asset.
+		/// Buyout amount given as input is the amount of native token that caller will receive in exchange for an amount of an allowed asset
+		/// Exchange amount given as input is the amount of an allowed asset that caller will give in exchange for an amount of native token
```
ebma, line 140:
```suggestion
		/// When denoting the `amount` as `Buyout` the caller will receive this exact amount of the native token in exchange for a corresponding amount of an allowed asset.
		/// When denoting the `amount` as `Exchange`, the caller will spend this exact amount of an allowed asset in exchange for a corresponding amount of the native token.
```

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -67,7 +68,7 @@ pub mod pallet {
 		type SellFee: Get<Permill>;
 
 		/// Type that allows for checking if currency type is ownable by users
-		type AllowedCurrencyIdVerifier: AllowedCurrencyIdVerifier<CurrencyIdOf<Self>>;
+		type AllowedCurrencyVerifier: AllowedCurrencyChecker<CurrencyIdOf<Self>>;
```
ebma, line 71:
```suggestion
		type AllowedCurrencyChecker: AllowedCurrencyChecker<CurrencyIdOf<Self>>;
```

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -241,8 +247,8 @@ impl<T: Config> Pallet<T> {
 		let (basic_asset_price, exchange_asset_price) = Self::fetch_prices((&basic_asset, &asset))?;
 
 		// Add fee to the basic asset price
-		let basic_asset_price_with_fee =
-			basic_asset_price * (FixedU128::from(T::SellFee::get()) + FixedU128::one());
+		let fee_plus_one = FixedU128::from(T::SellFee::get()).checked_add(&FixedU128::one()).expect("This should never fail");
```
ebma, line 250: Nit: Yes it should never fail. But let's not use `unwrap()` or `expect()` in any production code of our pallets. Tests and benchmarks can be treated differently. But if for some reason, the `SellFee` is the maximum possible value and we want to add `1` extra, this would make our runtime panic and halt our chain.
Even if there is no such scenario, we should still not use `unwrap()`/`expect()` anywhere. Instead please make this return a custom error instead. I'm fine if we have a generic name like `MathError` for this.

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -268,8 +274,8 @@ impl<T: Config> Pallet<T> {
...
```
ebma, line 277: Same here.

pallets/treasury-buyout-extension/src/mock.rs
```diff
@@ -148,7 +148,7 @@ impl orml_currencies::Config for Test {
 }
 
 pub struct AllowedCurrencyIdVerifierImpl;
```
ebma, line 150: Nit: probably makes sense to rename this as well

runtime/foucoco/src/lib.rs
```diff
@@ -996,7 +994,7 @@ impl orml_tokens_management_extension::Config for Runtime {
 }
 
 pub struct AllowedCurrencyIdVerifierImpl;
```
ebma, line 996: Same here

pallets/treasury-buyout-extension/src/lib.rs
ebma: Nice, thanks for explaining.

I'm also okay to leave it like this. We could rename some variables to make it a little clearer, e.g. we can rename `now` -> `current_block_number` and `current_period` -> `current_period_start_number`, because the name 'period' to me indicates some range but here we only calculate the starting block of that range.

runtime/foucoco/src/lib.rs
ebma: That's indeed a nice catch @gianfra-t 👍 
We certainly have some inaccuracy here, also considering that the call to `Security::set_status` will also result in an extra write operation. I agree with @bogdanS98 that it might be too much of a hassle to make this perfect because we'd have to introduce some coupling and the structure becomes more complex overall. 

However, we could at least think about adding an extra check at the beginning of this function, which checks if the `Oracle` already has a price for the requested `currency_id`. Then we can at least minimize the overhead of this function in case it gets called multiple times.

pallets/treasury-buyout-extension/src/lib.rs
bogdanS98: Completely agree! Going to change this now

runtime/foucoco/src/lib.rs
gianfra-t: That makes more sense. With that extra check only one of the 50(ish) samples would get benchmarked.

bogdanS98: Added the check and regenerated the weights 👍🏼

pallets/treasury-buyout-extension/src/lib.rs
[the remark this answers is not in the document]
bogdanS98: Since `SignedExtension` is an additional safety measure for pre-flight validation of the transactions (more like spam prevention) and might be bypassed somehow, I think it's important to also have those checks in the logic of the pallet.

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -0,0 +1,525 @@
...
+
+		let basic_asset = <T as orml_currencies::Config>::GetNativeCurrencyId::get();
+		let (buyout_amount, exchange_amount) = Self::split_to_buyout_and_exchange(asset, amount)?;
+
+		Self::ensure_buyout_limit_not_exceeded(&who, buyout_amount)?;
+		let treasury_account_id = T::TreasuryAccount::get();
+
+		// Start exchanging
+		// Check for exchanging zero values and same accounts
+		if exchange_amount.is_zero() && buyout_amount.is_zero() || who == treasury_account_id {
```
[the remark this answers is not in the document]
bogdanS98: Same as [here](https://github.com/pendulum-chain/pendulum/pull/396#discussion_r1480273783)

ebma (approved):
Just left some minor comments but I think it's good already. 

I also understand that the idea behind SignedExtensions is to have a mechanism to preemptively do a basic validity check for extrinsic calls so that invalid calls get filtered out early before they consume too much computing time. (Which is very important for fee-less extrinsics).

pallets/treasury-buyout-extension/src/lib.rs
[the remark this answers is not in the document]
gianfra-t: Quite dangerous to think that it is not optional. Thanks!

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -0,0 +1,543 @@
...
+		/// When attempting to buyout after this period, the buyout limit is reset for the caller
+		#[pallet::constant]
+		type BuyoutPeriod: Get<u32>;
+
+		/// Fee from the native asset buyouts
+		#[pallet::constant]
+		type SellFee: Get<Permill>;
+
+		/// Type that allows for checking if currency type is ownable by users
+		type AllowedCurrencyChecker: AllowedCurrencyChecker<CurrencyIdOf<Self>>;
```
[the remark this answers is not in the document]
bogdanS98: I think that is a great idea 👍🏼. Since we may open multiple new HRMP channels in the future, I think that in the long run we are going to save quite some time and effort by doing a root transaction instead of updating the runtime everytime we want to allow/deny an asset. Even if we don't change the allowed assets too often, in my opinion it's still worth storing the supported currencies.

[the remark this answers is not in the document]
bogdanS98: Added the changes for storing allowed currencies @ebma @gianfra-t

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -185,6 +221,58 @@ pub mod pallet {
...
+			for update in updates.clone() {
+				match update {
+					BuyoutAssetUpdate::Add(asset) => {
+						// Ensure native token is not added to the allowed assets for buyout
+						ensure!(asset != basic_asset, Error::<T>::NativeTokenNotAllowed);
+						AllowedCurrencies::<T>::insert(asset, ());
+						added_assets.push(Some(asset));
+					},
+					BuyoutAssetUpdate::Remove(asset) => {
+						AllowedCurrencies::<T>::remove(asset);
```
ebma, line 264: Does this throw if the asset doesn't exist yet?

ebma (approved):
Nice changes 👍 I think it would have been fine to just make the extrinsic such that the list is totally replaced, ie. implementing a setter and ignoring the last state. But it's fine either way 🤷

pallets/treasury-buyout-extension/src/lib.rs
bogdanS98: No, it doesn't throw any error if the asset doesn't exist. However, your suggestion would simplify the logic so I will make the list replaceable, ignoring the last state 👍🏼

bogdanS98: Done 👍🏼

pallets/treasury-buyout-extension/src/tests.rs
```diff
@@ -0,0 +1,642 @@
...
+		));
+
+		// Test user buyout after allowed currencies update
+		// It should succeed because dot is now allowed for buyout
+		assert_ok!(crate::Pallet::<Test>::buyout(
+			RuntimeOrigin::signed(user),
+			dot_currency_id,
+			Amount::Exchange(exchange_amount),
+		));
+	});
```
gianfra-t, line 245: Perhaps we could compare with the event (or storage directly, probably faster) to ensure that the allowed currencies were indeed modified to the expected `allowed_currencies` vec, WDYT?

gianfra-t (approved):
I checked the changes since last review and everything looks ready! I just left a comment but I don't think it is a blocker.

pallets/treasury-buyout-extension/src/tests.rs
bogdanS98: Done 👍🏼

ebma (commented):
I think it's good (better now with the simpler extrinsic). Just left a comment on the benchmark which is more an FYI, I don't mind if you don't change it @bogdanS98 and  a question.

pallets/treasury-buyout-extension/src/lib.rs
```diff
@@ -223,54 +222,56 @@ pub mod pallet {
...
-						added_assets.push(Some(asset));
-					},
-					BuyoutAssetUpdate::Remove(asset) => {
-						AllowedCurrencies::<T>::remove(asset);
-						removed_assets.push(Some(asset));
-					},
+			// Clear `AllowedCurrencies` storage
+			// `AllowedCurrencies` should have at most `max_allowed_currencies_for_buyout` entries
+			let result = AllowedCurrencies::<T>::clear(max_allowed_currencies_for_buyout, None);
+			// If storage clearing returns cursor which is `Some`, then clearing was not entirely successful
```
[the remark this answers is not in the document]
bogdanS98: `Some` is returned when not all the specified number of entries in the `StorageMap` are deleted in one call. In our case, the number of entries we want to delete in one call is up to `MaxAllowedBuyoutCurrencies` which we configure in the runtime. Given that this parameter should be equal to the number of all our supported currency ids in the worst case, then I don't think we have to worry about the storage not being fully cleared in one call since it's a small number of entries after all. We can use `u32::MAX` instead of `MaxAllowedBuyoutCurrencies` as a limit so we don't have to "trust" the correctness of the runtime parameter. Honestly, I don't know how we could end up in a situation where we can't reduce the amount of allowed currencies.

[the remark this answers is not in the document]
ebma: Ah true, we limit the number of items with `MaxAllowedBuyoutCurrencies` anyway. Well then we should be fine as long as we set that parameter to a reasonable number. Nevermind then.

ebma (approved):
I think it's ready for merge now 👍

ebma (approved):

PR #398, opened 2024-01-17
Add custom metadata and parameters for asset registry

### LINKED ISSUE #392: Add custom metadata and parameters for asset registry
We want to store a `fee_per_second` attribute for each asset defined in our asset registry. For this, we want to use the [CustomMetadata](https://github.com/pendulum-chain/pendulum/blob/54f466074384689e34d3e5af920884a230ea10a4/runtime/pendulum/src/lib.rs#L749) of the asset registry. At the moment, we are using the [CustomMetadata](https://github.com/pendulum-chain/pendulum/blob/54f466074384689e34d3e5af920884a230ea10a4/runtime/common/src/asset_registry.rs#L10) struct defined in the spacewalk primitives but it makes more sense to now define this struct in our runtimes instead.

## TODO
- Create a `CustomMetadata` struct that replaces the one in [asset_registry](https://github.com/pendulum-chain/pendulum/blob/54f466074384689e34d3e5af920884a230ea10a4/runtime/common/src/asset_registry.rs#L10)
- Add a `fee_per_second` parameter to it
- Add parameters for price oracle keys. We should create a new struct `DiaKeys { blockchain: Vec<u8>, symbol: Vec<u8> }` that allows using this info to create price info from our DIA oracle pallet.

bogdanS98:
Closes #392 .

### DIFF
--- a/runtime/amplitude/src/lib.rs
+++ b/runtime/amplitude/src/lib.rs
@@ -21,6 +21,7 @@ use zenlink_protocol::{AssetBalance, MultiAssetsHandler, PairInfo};
 pub use parachain_staking::InflationInfo;
 
 use codec::Encode;
+use scale_info::TypeInfo;
 
 use smallvec::smallvec;
 use sp_api::impl_runtime_apis;
@@ -830,6 +831,8 @@ impl orml_tokens::Config for Runtime {
 
 parameter_types! {
 	pub const NativeCurrencyId: CurrencyId = CurrencyId::Native;
+	#[derive(Clone, Eq, PartialEq, Debug, TypeInfo)]
+	pub const StringLimit: u32 = 50;
 }
 
 impl orml_currencies::Config for Runtime {
@@ -841,7 +844,7 @@ impl orml_currencies::Config for Runtime {
 
 impl orml_asset_registry::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
-	type CustomMetadata = asset_registry::CustomMetadata;
+	type CustomMetadata = asset_registry::CustomMetadata<StringLimit>;
 	type AssetId = CurrencyId;
 	type AuthorityOrigin = asset_registry::AssetAuthority;
 	type AssetProcessor = asset_registry::CustomAssetProcessor;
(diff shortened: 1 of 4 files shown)


runtime/amplitude/src/lib.rs
```diff
@@ -830,6 +831,8 @@ impl orml_tokens::Config for Runtime {
 
 parameter_types! {
 	pub const NativeCurrencyId: CurrencyId = CurrencyId::Native;
+	#[derive(Clone, Eq, PartialEq, Debug, TypeInfo)]
```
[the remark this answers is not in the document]
bogdanS98: It is needed because of the `AssetProcessor` implementation. `derive` indeed works inside `parameter_types!`

runtime/common/src/asset_registry.rs
```diff
@@ -4,21 +4,37 @@ use frame_system::EnsureRoot;
...
 use scale_info::TypeInfo;
-use sp_runtime::DispatchError;
+use sp_core::Get;
+use sp_runtime::{BoundedVec, DispatchError};
+use sp_std::fmt::Debug;
 use spacewalk_primitives::CurrencyId;
 
-pub use spacewalk_primitives::CustomMetadata;
+#[derive(Clone, PartialOrd, Ord, PartialEq, Eq, Debug, Encode, Decode, TypeInfo, MaxEncodedLen)]
+pub struct CustomMetadata<T: Get<u32> + TypeInfo + Clone + Eq + Debug + Send + Sync> {
```
gianfra-t, line 13: All seems good to me!. Just wondering on this line, why does `T` has to implement `TypeInfo`, `Clone`, etc?
I think this is related to what @ebma said [here](https://github.com/pendulum-chain/pendulum/pull/398/files#r1456080048), which is going to be required if we have these bounds. Is it because we are passing `T` to `DiaKeys<T>`?

gianfra-t (commented):
All good! just left one comment.

runtime/common/src/asset_registry.rs
bogdanS98: It is needed for the implementation of `AssetProcessor`

gianfra-t: Oh I see now. AssetProcessor also has those bounds.

runtime/amplitude/src/lib.rs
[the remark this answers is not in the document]
ebma: Alright 👍 I think it looks a bit ugly, but if it works, it works 😅

ebma (approved):
Looks good to me then

TorstenStueber (approved):

PR #401, opened 2024-01-25
addition of spwacewalk pallets to pendulum

### LINKED ISSUE #400: Implement Spacewalk on Pendulum
Original task [issue](https://github.com/pendulum-chain/tasks/issues/225).

This is the task ticket for the runtime upgrade containing: Spacewalk launch on Pendulum

#### Requirements:
We need to add all Spacewalk-related pallets to Pendulum. These are:

currency
fee
issue
redeem
nomination
oracle
replace
security
stellar-relay
vault-registry
pooled-rewards
vault-staking
clients-info
reward-distribution

The configuration of each pallet can be adopted from the respective configuration of Amplitude's runtime. Some things have to be adjusted though:

The DecayRate for the reward-distribution pallet should be Perquintill::from_parts(36600800000000000u64); according to [this](https://github.com/pendulum-chain/tasks/issues/47) ticket.

The implementation of the [DiaOracleKeyConvertor](https://github.com/pendulum-chain/pendulum/blob/2be0703b0c818bdd43fd32e2438eb4927d7f603c/runtime/amplitude/src/lib.rs#L186-L193) needs to be adjusted o work for Pendulum's assets. 

The native symbol/chain should be PEN and Pendulum (This has to match the identifiers of the PEN price known to the DIA oracle pallet) and the XCMCurrencyConversion needs to replace KSM with DOT and Kusama with Polkadot.

gianfra-t:
Please ensure that the collator nodes are redeployed after this PR is merged.  This is because of the new RPC methods that Spacewalk brings (and genesis config).

Closes #400 . 

### Changes

This PR is for the addition of all the relevant Spacewalk pallets into the Pendulum runtime. It mostly adopt the same configuration used in Amplitude, except for:

- [DiaOracleKeyConverter](https://github.com/pendulum-chain/pendulum/pull/401/files#diff-4598ea8bf56637ff9e107766864d676517486f61877f38c495a9a8e87c56da3dR161) converter, where chain and token is different.
- [Decay rate](https://github.com/pendulum-chain/pendulum/pull/401/files#diff-4598ea8bf56637ff9e107766864d676517486f61877f38c495a9a8e87c56da3dR1229) of the per-block rewards.

### Extra notes
- Although we must add entries into the genesis config for the new pallets, this values are taken from Amplitude as-is. In any case this values will not be used on the live chain.
- On storage versions: After addition, the new pallets will have a storage version of 0 or default. This is fine since none of them have ever performed migrations.


gianfra-t:
@ebma should we test the functionality of the newly added pallets in something like zombienet/chopsticks? Or is it okay since we already know they work in Amplitude? WDYT

runtime/pendulum/Cargo.toml
ebma, line 1: The indentation here is a bit unfortunate. Maybe we should find some tooling that also formats the `Cargo.toml` files as different IDEs seem to result in different formatting too often.

node/src/chain_spec.rs
```diff
@@ -921,6 +923,84 @@ fn pendulum_genesis(
...
+		fee: pendulum_runtime::FeeConfig {
+			issue_fee: FixedU128::checked_from_rational(15, 10000).unwrap(), // 0.15%
+			issue_griefing_collateral: FixedU128::checked_from_rational(5, 100000).unwrap(), // 0.005%
+			redeem_fee: FixedU128::checked_from_rational(5, 1000).unwrap(),  // 0.5%
+			premium_redeem_fee: FixedU128::checked_from_rational(5, 100).unwrap(), // 5%
+			punishment_fee: FixedU128::checked_from_rational(1, 10).unwrap(), // 10%
+			replace_griefing_collateral: FixedU128::checked_from_rational(1, 10).unwrap(), // 10%
+		},
+		nomination: pendulum_runtime::NominationConfig { is_nomination_enabled: false },
+		dia_oracle_module: amplitude_runtime::DiaOracleModuleConfig {
```
ebma, line 995:
```suggestion
		dia_oracle_module: pendulum_runtime::DiaOracleModuleConfig {
```

ebma (commented):
Looks good to me overall. CI failed though. Maybe it's fixed by my suggested code changes, not sure.

ebma:
> should we test the functionality of the newly added pallets in something like zombienet/chopsticks? Or is it okay since we already know they work in Amplitude? WDYT

I think we should test it but let's rather test the whole runtime when we are about to release. Meaning, I would rather test the runtime once we have the PR to bump spec and transaction versions.

runtime/pendulum/Cargo.toml
gianfra-t: 🙁. Actually, I may be to blame on this one. I wanted to maintain the same indentation and spaces as in Amplitude's `.toml`.

gianfra-t:
>  let's rather test the whole runtime when we are about to release. 
Makes sense!

runtime/pendulum/Cargo.toml
```diff
@@ -194,34 +211,67 @@ std = [
...
+    "module-redeem-rpc-runtime-api/std",
+    "module-replace-rpc-runtime-api/std",
+    "module-pallet-staking-rpc-runtime-api/std",
+    "module-vault-registry-rpc-runtime-api/std",
+    "spacewalk-primitives/std",
+
+    # custom libraries from pendulum
+	  "orml-currencies-allowance-extension/std",
+	  "parachain-staking/std",
+	  "vesting-manager/std",
```
ebma, line 245: Now we have three different levels of spacing here.

ebma (approved):
Seems like formatting .toml files with rustfmt is an open [issue](https://github.com/rust-lang/rustfmt/issues/4091) that still has some [developments](https://github.com/rust-lang/rustfmt/pull/5240) going on. Not sure if it's worth to look for other tooling here in the meantime or to deal with it manually for the time being. 

The changes look good to me overall and the CI passed so I think this is ready to merge. (Ideally after the indentation is streamlined but that's just nitpicking.)

runtime/pendulum/Cargo.toml
gianfra-t: It is so strange because in my editor I don't see the different indentation! Last time it happened the same, now I can see what you mean.

ebma (approved):
Thanks for fixing it 🙏

PR #405, opened 2024-01-26
bump spec version

### LINKED ISSUE #404: Bump versions for Pendulum runtime upgrade 12
We need to bump versions in order to prepare the next runtime upgrade on Pendulum. Related to https://github.com/pendulum-chain/tasks/issues/208. This time we should try only to bump the `spec_version` and not the `transaction_version`.

gianfra-t:
Closes #404. Required for performing Pendulum's runtime [upgrade](https://github.com/pendulum-chain/tasks/issues/208) 12.

We don't bump transaction version as only pallets were added.

### DIFF
--- a/runtime/pendulum/src/lib.rs
+++ b/runtime/pendulum/src/lib.rs
@@ -249,7 +249,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_version: 12,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version: 10,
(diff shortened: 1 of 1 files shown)


TorstenStueber (approved):

ebma:
Looking at what changed in the meantime, I see that the [rollout of the new metadata ](https://github.com/pendulum-chain/pendulum/commit/b7865600bbe5d4470de28e96540044f8900fe693)is part of the upgrade. Do you think we need to bump the transaction version for this change?

PR #410, opened 2024-02-06
393 use fee per second from asset metadata in xcm config

### LINKED ISSUE #393: Use `fee_per_second` from asset metadata in XCM config
With #392 we have a `fee_per_second` attribute available for each asset. We want to use this value to replace the `RelativeValue`s defined eg [here](https://github.com/pendulum-chain/pendulum/blob/84e6dd10928085ab0ee534150da3a190db170138/runtime/pendulum/src/xcm_config.rs#L120). 

An example of an implementation that uses the `fee_per_second` attribute defined in the asset registry to derive XCM fees can be found [here](https://github.com/centrifuge/centrifuge-chain/blob/976d63612c94773359b1d8de16f2cd1aeb5a79f6/runtime/common/src/xcm.rs#L33).

gianfra-t:
Closes #393. 

This PR will replace our previous implementation of the `RelativeValue` and calculation of the xcm fee by means of `WeightToFee`.

We will now use the parameter stored in the custom metadata `fee_per_second` in order to obtain the fee, by replacing altogether the previously implemented trader `TakeFirstAssetTrader` for `AssetRegistryTrader`. See the definition for the [former](https://github.com/paritytech/polkadot-sdk/blob/9e6298e7f773c61d384677e6eebfa8d7eee9d2ac/cumulus/primitives/utility/src/lib.rs#L111) and the [latter](https://github.com/open-web3-stack/open-runtime-module-library/blob/91942bcc134c5f073b3198d2eb9ebdec48d8547f/asset-registry/src/impls.rs#L80C12-L80C31).

Now, the fee is calculated as is defined in `FixedRateAssetRegistryTrader` [here](https://github.com/open-web3-stack/open-runtime-module-library/blob/91942bcc134c5f073b3198d2eb9ebdec48d8547f/asset-registry/src/impls.rs#L55), where: `fee = weight/WEIGHT_REF_TIME_PER_SECOND * fee_per_second`

Another important difference is that `AssetRegistryTrader` will attempt to buy from all of the assets exchanged until one is able to do so, while `TakeFirstAsset` trader will only attempt the first. 

Also, if the asset by location is not found in the registry, the trader will error with `TooExpensive`, this is why [in the incorrect incoming asset test](https://github.com/pendulum-chain/pendulum/pull/410/files#diff-3429c5361e6214779992604d4212c994d6c327c55b74c9bcb280681ad5ddf93dR245) we now look for this error.

The `FixedConversionRateProvider` type that  `FixedRateAssetRegistryTrader` requires will just [attempt](https://github.com/pendulum-chain/pendulum/pull/410/files#diff-3287a4ee060b1070a8e6559a094b4890a596a1b281effef54cafe59da20edf49R86) to get the `fee_per_second` from the metadata. 

Finally, we now require to [declare](https://github.com/pendulum-chain/pendulum/pull/410/files#diff-dffe390f73891278c96aff2bbe6d1a6146bddc57ce4c8621e0a15aeae822e0c1R225) the metadata in our mock chains for integration tests, particularly the `fee_per_second` value, so that the message is able to execute. For this, we only care that the correct multilocation of the incoming asset exists in the registry, with the corresponding `fee_per_second`. For simplicity, this value is the same for each asset.


runtime/common/src/lib.rs
```diff
@@ -70,26 +73,19 @@ pub mod opaque {
...
-		if let Some(adjusted_amount) = amount
-			.clone()
-			.saturating_mul(relative_value.denominator)
-			.checked_div(&relative_value.num)
-		{
-			return adjusted_amount
-		}
-		// We should never specify a numerator of 0, but just to be safe
-		return amount
+pub struct FixedConversionRateProvider<OrmlAssetRegistry>(PhantomData<OrmlAssetRegistry>);
```
bogdanS98, line 76: Maybe it's better to have this in `asset_registry.rs` instead? WDYT?

bogdanS98 (commented):
Looks good to me 👍🏼 I left only one comment. We still need to verify if the checks are successful once #396 is merged into main, and subsequently after main is merged into this one.

runtime/common/src/lib.rs
gianfra-t: Yes I agree! moving.

runtime/common/src/asset_registry.rs
```diff
@@ -54,3 +56,20 @@ impl AssetProcessor<CurrencyId, AssetMetadata<Balance, CustomMetadata>> for Cust
...
+		OrmlAssetRegistry: Inspect<
+			AssetId = CurrencyId,
+			Balance = Balance,
+			CustomMetadata = asset_registry::CustomMetadata,
+		>,
+	> FixedConversionRateProviderTrait for FixedConversionRateProvider<OrmlAssetRegistry>
+{
+	fn get_fee_per_second(location: &MultiLocation) -> Option<u128> {
+		let metadata = OrmlAssetRegistry::metadata_by_location(location)?;
+		Some(metadata.additional.fee_per_second)
```
ebma, line 73: It might be best if we try to come up with a default fee in case there is no fee stored for the asset in the registry. Something similar to [this](https://github.com/centrifuge/centrifuge-chain/blob/976d63612c94773359b1d8de16f2cd1aeb5a79f6/runtime/common/src/xcm.rs#L48).

runtime/common/src/lib.rs
```diff
@@ -2,10 +2,12 @@
 #![allow(non_snake_case)]
 
 use sp_runtime::{
-	traits::{CheckedDiv, IdentifyAccount, Saturating, Verify},
+	traits::{IdentifyAccount, Verify},
 	DispatchError, MultiSignature,
 };
 
+use spacewalk_primitives::CurrencyId;
+
```
ebma, line 10: Do we need this import?

runtime/integration-tests/src/mock.rs
```diff
@@ -226,27 +408,31 @@ impl Builder<PendulumCurrencyId> for ExtBuilderParachain<PendulumCurrencyId> {
...
 				use amplitude_runtime::{Runtime, System};
-				build_parachain_with_orml!(
+				build_parachain_with_orml_and_asset_registry!(
 					self,
 					Runtime,
 					System,
 					NATIVE_INITIAL_BALANCE,
 					ORML_INITIAL_BALANCE,
-					PendulumCurrencyId
+					PendulumCurrencyId,
```
ebma, line 434: This issue existed before but I assume it should be (?)
```suggestion
					AmplitudeCurrencyId,
```

ebma (commented):
Looks good overall 👍  Few remarks:
- It might be more interesting not to use the same `fee_per_second` value for each asset in the test
- We should also change the xcm config for Foucoco
- We might want to introduce a default value in case the `fee_per_second` is not available

runtime/common/src/asset_registry.rs
gianfra-t: I saw that example but wasn't sure if it was best to define a default that may be insecure given a possible low token value, or simply do not allow to pay fee on that asset, which is what I decided in the end. Because I see that in the example they define a base fee based on the decimals, but this still makes no assumption of the possible value of that token.

Any idea how can we define a default that we can ensure is safe, even with low value assets?

runtime/common/src/lib.rs
gianfra-t: Nope! removing

runtime/integration-tests/src/mock.rs
gianfra-t: Yes, although they are the same type, it makes sense to maintain naming convention.

gianfra-t:
@ebma thanks for pointing out about the xcm config of Foucoco. I didn't consider it. We should keep in mind that any XCM message was allowed before, given the configuration of the barrier `AllowUnpaidExecutionFrom<Everything>`. Do you think this could break any integration we have right now? For example, with moonbase alpha.

runtime/common/src/asset_registry.rs
ebma: As discussed in the sync, I agree it's best to return with an error here and not implement a default `fee_per_second`. In practice, the `fee_per_second` should always be available for each asset we want to support.

gianfra-t: Allright! The error would eventually be returned [here](https://github.com/open-web3-stack/open-runtime-module-library/blob/91942bcc134c5f073b3198d2eb9ebdec48d8547f/asset-registry/src/impls.rs#L108), given that the fee would not be found, and it will just return `TooExpensive`. I do not know how a different error could be returned without modifying `AssetRegstiryTrader`, but if you think it is relevant we can implement it locally and modify that.

ebma:
Good question. I doubt that it messes with any of our Foucoco integrations (at the moment just Moonbase Alpha). Should be fine to change it like this. Or what do you think @bogdanS98?

bogdanS98:
> Good question. I doubt that it messes with any of our Foucoco integrations (at the moment just Moonbase Alpha). Should be fine to change it like this. Or what do you think @bogdanS98?

I think it's fine as long as we have the same XCM config for all runtimes with respect to `Barrier`, `Trader` and `AssetTransactor`. This PR gets all configurations on the same page and I can't think of any reserved transfer scenario that would be different in Foucoco than in the other runtimes.

gianfra-t:
That is great to hear @bogdanS98 @ebma. It is nice indeed to have all the same XCM configurations on the chains otherwise testing becomes very difficult. I think I will wait a bit [this](https://github.com/pendulum-chain/pendulum/issues/416) discussion and then, if you agree with the changes, we can merge.

node/src/constants/foucoco.rs
```diff
@@ -9,6 +9,8 @@ pub const INITIAL_COLLATOR_STAKING: Balance = 10_000 * UNIT;
 pub const COLLATOR_ADDITIONAL: Balance = 10 * UNIT;
 
 pub const OFF_CHAIN_WORKER_ADDRESS: &str = "6m69vWMouLarYCbJGJisVaDDpfNGETkD5hsDWf2T7osW4Cn1";
+pub const ALICIA: &str = "6mfqoTMHrMeVMyKwjqomUjVomPMJ4AjdCm1VReFtk7Be8wqr";
```
ebma, line 12: What is `ALICIA` used for?

runtime/common/src/asset_registry.rs
ebma: Yes, sorry the `TooExpensive` error is enough. I was referring to that but phrased it poorly, so I'm fine if we keep this as is 👍

node/src/constants/foucoco.rs
gianfra-t: It's a leftover from a test. Deleting!

node/src/chain_spec.rs
```diff
@@ -204,8 +206,8 @@ pub fn foucoco_config() -> FoucocoChainSpec {
 				// initial collators.
 				invulnerables.clone(),
 				signatories.clone(),
-				vec![sudo_account.clone(), offchain_worker_price_feeder.clone()],
-				sudo_account.clone(),
+				vec![sudo_account.clone(), offchain_worker_price_feeder.clone(), alicia.clone()],
+				alicia.clone(),
```
ebma, line 210: Oh wait, we need to change this too, right?

runtime/integration-tests/src/amplitude_tests.rs
```diff
@@ -16,9 +16,11 @@ use statemine_runtime as kusama_asset_hub_runtime;
 use xcm::latest::NetworkId;
 use xcm_emulator::{decl_test_network, decl_test_parachain, decl_test_relay_chain, TestExt};
 
-const KSM_FEE_WHEN_TRANSFER_TO_PARACHAIN: polkadot_core_primitives::Balance = 32000000;
-const USDT_FEE_WHEN_TRANSFER_TO_PARACHAIN: polkadot_core_primitives::Balance = 640000000;
-const NATIVE_FEE_WHEN_TRANSFER_TO_PARACHAIN: polkadot_core_primitives::Balance = 3200000000;
+// Since in the mock definition all the assets are defined with the same fee_per_second value
+// the total fees for every asset will be the same.
```
ebma, line 20: We might want to remove this comment again.

runtime/integration-tests/src/pendulum_tests.rs
```diff
@@ -18,10 +18,12 @@ use statemint_runtime as polkadot_asset_hub_runtime;
 use xcm::latest::NetworkId;
 use xcm_emulator::{decl_test_network, decl_test_parachain, decl_test_relay_chain, TestExt};
 
-const DOT_FEE_WHEN_TRANSFER_TO_PARACHAIN: polkadot_core_primitives::Balance = 32653061; //The fees that relay chain will charge when transfer DOT to parachain. sovereign account of some parachain will receive transfer_amount - DOT_FEE
-const MOONBEAM_BRZ_FEE_WHEN_TRANSFER_TO_PARACHAIN: polkadot_core_primitives::Balance = 1391304347; //Fees that we will charge in incoming Moonbeam's BRZ. Depends on the RelativeValue struct implementation.
-const USDT_FEE_WHEN_TRANSFER_TO_PARACHAIN: polkadot_core_primitives::Balance = 266666666;
-const NATIVE_FEE_WHEN_TRANSFER_TO_PARACHAIN: polkadot_core_primitives::Balance = 3200000000;
+// Since in the mock definition all the assets are defined with the same fee_per_second value
+// the total fees for every asset will be the same.
```
ebma, line 22: Same here

runtime/integration-tests/src/amplitude_tests.rs
```diff
@@ -16,9 +16,11 @@ use statemine_runtime as kusama_asset_hub_runtime;
...
+const KSM_FEE_WHEN_TRANSFER_TO_PARACHAIN: polkadot_core_primitives::Balance = 200000000; // /20 native token value fee per second
+const USDT_FEE_WHEN_TRANSFER_TO_PARACHAIN: polkadot_core_primitives::Balance = 400000000;// /10 native token value fee per second
+const NATIVE_FEE_WHEN_TRANSFER_TO_PARACHAIN: polkadot_core_primitives::Balance = 4000000000;
```
ebma, line 23: Maybe we should define these constants relative to each other, ie. change it to something like
```
// According to the `fee_per_second` defined in the mock
const NATIVE_FEE_WHEN_TRANSFER_TO_PARACHAIN: polkadot_core_primitives::Balance = 4000000000;
const KSM_FEE_WHEN_TRANSFER_TO_PARACHAIN: polkadot_core_primitives::Balance = NATIVE_FEE_WHEN_TRANSFER_TO_PARACHAIN / 20;
const USDT_FEE_WHEN_TRANSFER_TO_PARACHAIN: polkadot_core_primitives::Balance = NATIVE_FEE_WHEN_TRANSFER_TO_PARACHAIN / 10
```

so we don't need the comments.

runtime/integration-tests/src/pendulum_tests.rs
```diff
@@ -18,10 +18,12 @@ use statemint_runtime as polkadot_asset_hub_runtime;
...
+const DOT_FEE_WHEN_TRANSFER_TO_PARACHAIN: polkadot_core_primitives::Balance = 1000000000; // /4 native token value fee per second 
+const MOONBEAM_BRZ_FEE_WHEN_TRANSFER_TO_PARACHAIN: polkadot_core_primitives::Balance = 8000000000; // 2x native token value fee per second
+const USDT_FEE_WHEN_TRANSFER_TO_PARACHAIN: polkadot_core_primitives::Balance = 2000000000; // /2 native token value fee per second
+const NATIVE_FEE_WHEN_TRANSFER_TO_PARACHAIN: polkadot_core_primitives::Balance = 4000000000;
```
ebma, line 26: Same here

ebma (approved):
I just commented some minor things about the comments. I think we can improve it a little but it's not important.

runtime/integration-tests/src/amplitude_tests.rs
gianfra-t: Yes!

ebma (approved):

PR #413, opened 2024-02-16
Fix `ConvertPrice` implementations

### LINKED ISSUE #412: Convert DIA prices to correct FixedPoint numbers
## Context 

At the moment, the conversion of `u128` prices to `UnsignedFixedPoint` numbers we use in our runtimes (see [here](https://github.com/pendulum-chain/pendulum/blob/cbe46145223c9c660304ebc1ea2306d27fe45501/runtime/foucoco/src/lib.rs#L209)) is not correct. 

The prices are 10 ** 6 times smaller than they should because we are using `u128` as the `inner` type of our `UnsignedFixedPoint` implementation, see [here](https://github.com/pendulum-chain/spacewalk/blob/82f9e432490b85234f16538b996c589e6af4146e/primitives/src/lib.rs#L374),  which makes the FixedPoint implementation use a `DIV` of `1_000_000_000_000_000_000` (10e18), see [here](https://github.com/paritytech/substrate/blob/98f2e3451c9143278ec53c6718940aeabcd3b68a/primitives/arithmetic/src/fixed_point.rs#L2210). The batching server converts the prices to `u128` numbers considering 12 decimals in [this](https://github.com/pendulum-chain/oracle-pallet/blob/716073885de01f923a0fe44a05bd2a0bd45db555/dia-batching-server/src/price_updater.rs#L141) function however. Thus we have a mismatch.

This is not problematic for the price conversions in Spacewalk but we should still try to fix this.

Related notion page [here](https://www.notion.so/satoshipay/Spacewalk-price-conversion-accuracy-c3ebac43432c403fbdace469f400e59f). 

## TODO
Change the conversion functions in the runtime to convert the incoming prices to correct FixedPoint numbers.

ebma:
Instead of using `from_inner()` we create the fixed point number with `from_rational()` and divide it by 10e12 since we know that these are the decimals considered by the batching server. 

Closes #412.

### DIFF
--- a/runtime/amplitude/src/lib.rs
+++ b/runtime/amplitude/src/lib.rs
@@ -196,7 +196,9 @@ pub struct ConvertPrice;
 
 impl Convert<u128, Option<UnsignedFixedPoint>> for ConvertPrice {
 	fn convert(price: u128) -> Option<UnsignedFixedPoint> {
-		Some(UnsignedFixedPoint::from_inner(price))
+		// The DIA batching server returns the price in 1e12 format, see [here](https://github.com/pendulum-chain/oracle-pallet/blob/716073885de01f923a0fe44a05bd2a0bd45db555/dia-batching-server/src/price_updater.rs#L141)
+		// but our UnsignedFixedPoint implementation expects the price in 1e18 format.
+		Some(UnsignedFixedPoint::from_rational(price, 1_000_000_000_000))
 	}
 }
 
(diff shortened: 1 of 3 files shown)


gianfra-t (approved):
I think this is great! I was very confused the other day with respect to the other [fix](https://github.com/pendulum-chain/spacewalk/pull/490), and although this wasn't breaking anything, it is way better to be consistent in the intermediate representation of the prices.

bogdanS98 (approved):
Looks good to me too 👍🏼

PR #426, opened 2024-02-25
Use our hotfixed version of pallet-contracts

TorstenStueber:
Don't merge this PR! This is only to be used as a reference about how the current Foucoco has been built. We can close this PR (without merging), once we upgraded to Polkadot version v0.9.42.


TorstenStueber:
We upgraded to Polkadot version v0.9.42 now and therefore can close this here.

PR #429, opened 2024-03-01
Register Pink token on Pendulum

bogdanS98:
Closes [#tasks/issues/254](https://github.com/pendulum-chain/tasks/issues/254).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -191,6 +191,7 @@ version = "0.1.0"
 dependencies = [
  "bifrost-farming",
  "bifrost-farming-rpc-runtime-api",
+ "cfg-if",
  "clients-info",
  "cumulus-pallet-aura-ext",
  "cumulus-pallet-dmp-queue",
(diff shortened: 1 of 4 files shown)


ebma (changes requested):
I think we also need to register it [here](https://github.com/pendulum-chain/pendulum/blob/570a41a13e14f1b96fe2bb2fb6e81a49681c14a8/runtime/pendulum/src/xcm_config.rs#L118) because it's an XCM asset which otherwise would not be transferrable to us.

bogdanS98:
> I think we also need to register it [here](https://github.com/pendulum-chain/pendulum/blob/570a41a13e14f1b96fe2bb2fb6e81a49681c14a8/runtime/pendulum/src/xcm_config.rs#L118) because it's an XCM asset which otherwise would not be transferrable to us.

I merged main into my branch and now I have @gianfra-t's [changes](https://github.com/pendulum-chain/pendulum/pull/410) regarding XCM fees so we don't need this anymore

ebma (approved):
LGTM

PR #430, opened 2024-03-12
170 Call Gitlab CI job to Automate Runtime Release

b-yap:
partially closes https://github.com/pendulum-chain/tasks/issues/170.

* ~~added a script for picking commits related to specific runtime;~~
* ~~added a github action for srtool (based on [bifrost](https://github.com/bifrost-finance/bifrost/blob/develop/.github/workflows/srtool.yml)'s)~~
* ~~added a js for authorizing upgrade based on the srtool.~~ 

~~[Full info can be found in the issue](https://github.com/pendulum-chain/tasks/issues/170#issuecomment-2009313436).~~


.github/workflows/release.yml
```diff
@@ -0,0 +1,136 @@
...
+on:
+  pull_request:
+    types:
+      - closed
+    branches:
+      - 'main'
+
+jobs:
+  srtool:
+    if: ${{ github.event.pull_request.merged == true && contains(github.event.pull_request.title, 'release:') }}
```
b-yap, line 15: This job will only run if:
*  the pull request is closed and merged to main branch;
*  the pull request has the _**"release:"**_ in its title

.github/workflows/release.yml
```diff
@@ -0,0 +1,136 @@
...
+      matrix:
+        chain: ["amplitude", "pendulum", "foucoco"]
+        shouldReleaseAmp:
+          - ${{ contains(github.event.pull_request.title, 'amplitude') }}
+        shouldReleasePen:
+          - ${{ contains(github.event.pull_request.title, 'pendulum') }}
+        shouldReleaseFou:
+          - ${{ contains(github.event.pull_request.title, 'foucoco') }}
+        exclude:
+          - shouldReleaseAmp: false
```
b-yap, line 28: if the pull request title does not have the chain (foucoco, amplitude, or pendulum) in the title, it will be removed in the list.

.github/workflows/release.yml
```diff
@@ -0,0 +1,136 @@
...
+          runtime_dir: runtime/${{ matrix.chain }}
+          # this is important to avoid build fail. See https://github.com/paritytech/srtool/issues/62
+          tag: 1.66.1
+      - name: Summary
+        run: |
+          echo '${{ steps.srtool_build.outputs.json }}' | jq > ${{ matrix.chain }}-srtool-digest.json
+          cat ${{ matrix.chain }}-srtool-digest.json
+          echo "Runtime location: ${{ steps.srtool_build.outputs.wasm }}"
+          echo '${{ steps.srtool_build.outputs.json }}'
+          cp ${{ steps.srtool_build.outputs.wasm }} ${{ matrix.chain }}_runtime.compact.wasm
```
b-yap, line 56: this will copy the file found deeply nested in directories, to the current directory.

.github/workflows/release.yml
```diff
@@ -0,0 +1,136 @@
...
+        with:
+          name: ${{ matrix.chain }}-runtime-${{ github.sha }}
+          path: |
+            ${{ matrix.chain }}-info.json
+            ${{ matrix.chain }}-info_compressed.json
+            ${{ matrix.chain }}-metadata.json
+            ${{ matrix.chain }}-diff.txt
+
+  propose_upgrade:
+    if: ${{ success() }}
```
b-yap, line 96: this job only runs if the srtool job ended successfully.

.github/workflows/release.yml
```diff
@@ -0,0 +1,136 @@
...
+    needs: srtool
+    name: Perform runtime upgrade on ${{ matrix.chain }} ${{ github.event.inputs.ref }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        chain: ["amplitude", "pendulum"]
```
b-yap, line 102: we do not include foucoco on this release procedure.

.github/workflows/release.yml
```diff
@@ -0,0 +1,136 @@
...
+          node-version: '21.x'
+      - name: Install dependencies
+        run: |
+          cd .github/scripts/js
+          npm init -y
+          npm install
+      - name: run script for parachain upgrade
+        run: |
+          node --version
+          node .github/scripts/js/parachain_authorize_upgrade.js ${{ matrix.chain }} ${{ matrix.chain }}_runtime.compact.compressed.wasm
```
b-yap, line 136: calls the script to perform the proposal transaction

.github/workflows/scripts/js/common.js
```diff
@@ -0,0 +1,346 @@
...
+      }
+
+      if (status.isFinalized) {
+        console.log("Transaction finalized");
+      }
+    });
+  });
+}
+
+async function democracyProposal(call, type, deposit, submitPreimage, { api, submitterKeypair, dryRun }) {
```
b-yap, line 54: I have commented all `submitExtrinsic` functions, until we're ready to call them.

.github/workflows/scripts/get_commits.sh
```diff
@@ -0,0 +1,62 @@
...
+latestCommitOfLastVersion=$(git log $lastVersionName --oneline --max-count=1 | cut -c1-7)
+
+if [ -z "${latestCommitOfLastVersion}" ]; then
+  echo "last version is up to date."
+  exit
+fi
+
+## only list commits related to this runtime and the general code changes
+## save to file commits_for_<chosen_runtime>-release-<new version number>.txt
+git log $latestCommitOfLastVersion..origin --oneline | grep -i -Ev "$excludeRuntimes" |cut -c1-7 >> $fileName
```
gianfra-t, line 60: I encountered in local testing that had to put `origin/main`, it is probably different in the workflows right? I imagine it works but not sure why locally it doesn't.

gianfra-t (commented):
Looks very good @b-yap !! This will save us a lot of time and potential errors.
I only have one comment but it may not be feasible. Since `common.js` and the constants are taken from the transaction builder repo, do you think it makes sense to fetch those file as one more step in the last work? So we don't have to keep track of both of them.

I think we could just use `run: curl ...` but then I don't know because the repo is private.

.github/workflows/scripts/js/parachain_authorize_upgrade.js
```diff
@@ -0,0 +1,44 @@
...
+const { ApiPromise, WsProvider, Keyring } = require("@polkadot/api");
+const { blake2AsHex } = require("@polkadot/util-crypto");
+const fs = require('fs');
+
+const { submitTransaction } = require("./common");
+
+const { getDefinitions } = require("./constants");
+
+const GOVERNANCE_MODE = "democracy"; // "democracy" | "sudo" | "multisig"
+const DRY_RUN = true;
```
ebma, line 16:
```suggestion
const DRY_RUN = false;
```

ebma (commented):
Wow really cool stuff @b-yap, very impressive! I think for the `.js` stuff, we should strip this down and reduce the files to maybe just one as there is plenty of code we don't need. Some other changes are required to make this work in CI but it already looks great overall 🙏

.github/workflows/scripts/js/common.js
```diff
@@ -0,0 +1,346 @@
...
+  console.log("Preimage data", call.inner.toHex());
+  console.log("Preimage hash", `0x${Buffer.from(call.inner.hash).toString("hex")}`);
+
+  if (submitPreimage) {
+    const submitPreimageTransaction = api.tx.preimage.notePreimage(call.inner.toHex());
+    // await submitExtrinsic(submitPreimageTransaction, submitterKeypair, dryRun);
+  }
```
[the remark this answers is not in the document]
b-yap: @ebma whose account? Should we bring this up later?

.github/workflows/release.yml
```diff
@@ -0,0 +1,136 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.inputs.ref }}
+          fetch-depth: 0
+
+      - name: download artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{ matrix.chain }}-runtime
```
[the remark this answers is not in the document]
b-yap: that is the name of the zip file, not the wasm file itself.

.github/workflows/scripts/js/common.js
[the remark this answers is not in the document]
b-yap: @ebma I asked @TorstenStueber and we're going to create different accounts for both pendulum and amplitude. 
This would mean calling this workflow has to be on gitlab, to secure the private key.

.github/workflows/scripts/get_commits.sh
b-yap: Can you share what the error looks like?

.github/workflows/scripts/js/common.js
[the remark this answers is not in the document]
ebma: Sounds good 👍

.github/workflows/release.yml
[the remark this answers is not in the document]
ebma: Which zip file? Where is it created?

.github/workflows/release.yml
```diff
@@ -0,0 +1,136 @@
...
+      - closed
+    branches:
+      - 'main'
+
+jobs:
+  srtool:
+    if: ${{ github.event.pull_request.merged == true && contains(github.event.pull_request.title, 'release:') }}
+    name: SrTool check on ${{ matrix.chain }} ${{ github.event.inputs.ref }}
+    strategy:
+      fail-fast: false
```
bogdanS98, line 18: Shouldn't this be true?

.github/workflows/release.yml
bogdanS98: Is there any way that srtool job can end with a "false positive" success?

.github/workflows/scripts/get_commits.sh
bogdanS98: For me the command works with both `origin` and `origin/main` locally, but I agree with @gianfra-t that we should be precise and specify the branch too.

bogdanS98 (commented):
This looks great, good job @b-yap 👍🏼 Can't wait to see it in action! I only left some questions.

.github/workflows/scripts/get_commits.sh
gianfra-t: Sorry I came back with a late reply, It now works from me! Must have been a local issue so never mind my comment.

.github/workflows/release.yml
b-yap: So this logic will be moved to gitlab, and the flow will look a bit different.

.github/workflows/scripts/get_commits.sh
b-yap: Noted. I've edited it to:
`git log $latestCommitOfLastVersion..origin/main `

.github/workflows/release.yml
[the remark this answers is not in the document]
b-yap: Torsten is correct. And the [@actions/download-artifact](https://github.com/actions/download-artifact?tab=readme-ov-file) will unzip it and extract the file.

Although the logic will look different in gitlab now.

b-yap:
All the scripts have been moved to gitlab.

.github/workflows/release-old.yml
```diff
@@ -0,0 +1,136 @@
+name: Release and Propose an Upgrade
```
b-yap, line 1: I'll remove this once everything's settled.

ebma (approved):
Looks good to me, though I did not check the respective commands of the job in Gitlab yet.

ebma:
I also checked the Gitlab code now and it all looks good to me. Let's merge this now or is there a reason not to do that yet @b-yap?

b-yap:
@ebma, I "resolved" the conversations but the `Squash and merge` button is still disabled. Can you try to "resolve" them and see if it works?

ebma:
@b-yap should work now, at least it's all green for me.

PR #431, opened 2024-03-14
Revert "393 use fee per second from asset metadata in xcm config"

ebma:
Reverts pendulum-chain/pendulum#410.

In order not to block the higher priority changes for the dependency upgrade to v0.9.42 and the PINK token registration, we'll roll back the changes from the main branch for now and merge them again at a later point in time. As far as I can tell this is the only thing that needs to get rolled back.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -191,7 +191,6 @@ version = "0.1.0"
 dependencies = [
  "bifrost-farming",
  "bifrost-farming-rpc-runtime-api",
- "cfg-if",
  "clients-info",
  "cumulus-pallet-aura-ext",
  "cumulus-pallet-dmp-queue",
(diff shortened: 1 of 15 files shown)


gianfra-t (approved):
Looks good! It was indeed only `a402b6204cbcb8cac5bb4d2801628f8a91bae683` that had to be removed.

TorstenStueber (commented):
I checked the output of git diff 4dda8b62489b553ace5a4ae536d240c0f7f24fd0 (the commit before merging #393) 5f878f06e827b55ad69a70b72d08088d375eaf1f (the commit of this PR) and the only changes are indeed only the changes of the two PRs that happened in between: #428 and #429.

So, looks good to me too.

TorstenStueber (approved):

PR #432, opened 2024-03-14
393 use fee per second from asset metadata in xcm config

### LINKED ISSUE #393: Use `fee_per_second` from asset metadata in XCM config

ebma:
This PR adds changes that were reverted before. The original PR is #410. 

We only want to merge this **after the dependency upgrade to v0.9.42** is merged _and released_. 

### TODO 
- [x] We still need to change the conversion function for `MultiLocation` <> `CurrencyId`. Currently, we are still using the hard-coded logic that manually maps these two but we also want to use the asset registry for this so no further runtime upgrades are needed for adding assets. Thus, we should change the conversion function similar to what is done [here](https://github.com/centrifuge/centrifuge-chain/blob/47fff268fc7cea828b034a3626df5cc289459a52/runtime/common/src/xcm.rs#L154-L202). 

Closes #393.


runtime/amplitude/src/xcm_config.rs
```diff
@@ -80,30 +65,32 @@ pub struct CurrencyIdConvert;
...
-			CurrencyId::XCM(index) => match index {
-				xcm_assets::RELAY_KSM => Some(MultiLocation::parent()),
-				xcm_assets::ASSETHUB_USDT => Some(asset_hub::USDT_location()),
-				_ => None,
-			},
-			CurrencyId::Native => Some(native_location_external_pov()),
-			_ => None,
-		}
+		AssetRegistry::metadata(&id)
+			.filter(|m| m.location.is_some())
```
gianfra-t, line 69: If there exist a location, we assume the asset is transferable. More specifically, that we allow it for paying fees.

runtime/common/src/asset_registry.rs
```diff
@@ -54,3 +56,24 @@ impl AssetProcessor<CurrencyId, AssetMetadata<Balance, CustomMetadata>> for Cust
...
+		OrmlAssetRegistry: Inspect<
+			AssetId = CurrencyId,
+			Balance = Balance,
+			CustomMetadata = asset_registry::CustomMetadata,
+		>,
+		CurrencyIdConvert: Convert<MultiLocation, Option<CurrencyId>>,
+	> FixedConversionRateProviderTrait for FixedConversionRateProvider<OrmlAssetRegistry, CurrencyIdConvert>
+{
+	fn get_fee_per_second(location: &MultiLocation) -> Option<u128> {
+		let asset_id_maybe = CurrencyIdConvert::convert(location.clone());
```
gianfra-t, line 71: Instead of looking the asset registry by `MultiLocation` we look the metadata by `CurrencyId`. Although this is the same in most cases, it removes the need to handle different `MultiLocation` representations for the same asset, which is a problem since we can only store one on the asset registry.
We therefore will delegate this to the `CurrencyIdConvert` which will normalize it as unanchored.

Example: we need to handle Native, which is both `{parent: 0, interior: PalletId(10)}`, `{parent: 1, interior: ParachainId(id), PalletId(10)}`

runtime/foucoco/src/xcm_config.rs
```diff
@@ -64,29 +66,31 @@ pub struct CurrencyIdConvert;
...
+                        interior: new_interior,
+                    },
+                    Err(_) => return None, 
+                }
+            },
+            x => x,
+        };
+		AssetRegistry::location_to_asset_id(unanchored_location)
 	}
 }
```
ebma, line 95: Can we try to generify this struct and put it into `runtime/common`? I think the only problem is the `ParachainInfo::parachain_id()` that is specific per runtime but we might be able to supply that to the struct instead and then we can have just one struct for all runtimes.

runtime/foucoco/src/xcm_config.rs
```diff
@@ -64,29 +66,31 @@ pub struct CurrencyIdConvert;
...
-			_ => None,
-		}
+		AssetRegistry::metadata(&id)
+			.filter(|m| m.location.is_some())
+			.and_then(|m| m.location)
+			.and_then(|l| l.try_into().ok())
 	}
 }
 
 impl Convert<MultiLocation, Option<CurrencyId>> for CurrencyIdConvert {
```
ebma, line 76: We might also want to have a comment similar to [this](https://github.com/centrifuge/centrifuge-chain/blob/47fff268fc7cea828b034a3626df5cc289459a52/runtime/common/src/xcm.rs#L176-L179) here so that it's clear why we are unanchoring the location. 

Actually, I don't think we even need to do this. In Pendulum's asset registry we put 'native' and the Stellar assets as `location: V2: {parents: 0, ... }`. And as far as I understand the code here, this would re-anchor the location from `parents: 0` to `parents: 1`, thus we wouldn't be able to find the asset in the registry anymore. Or am I misunderstanding the logic here? 

We could instead of course change the definitions in the asset registry so that every location is depicted as `parents: 1, ...` but we need to keep this in mind.

runtime/foucoco/src/xcm_config.rs
gianfra-t: Yes! We can leave `ParachainInfo` as generic. Good idea.

runtime/foucoco/src/xcm_config.rs
gianfra-t: Yes you got the logic right, and this is a decision we need to make. I assumed (incorrectly) that we were going to add all assets from the relay perspective, that is unanchored. Then as you can also see [here](https://github.com/pendulum-chain/pendulum/pull/432/files#diff-5dc9ab00db16dcf58b342b3f2a9ddd05a114370b9d04b514a3812083382b416dR71) we are converting now the location to currency before fetching the fee. This is mostly because I encountered the problem during testing that the anchored location of `Native` was being used.

But besides tests, I thought it is a good idea to be able to handle both representations. For instance, part of the message may execute from our chain and the representation in this case may be the anchored one.

Since we are bounded by other chains sending assets represented as unanchored, I think that is also the best way to represent our assets (Native and as you mention, Spacewalk). But I don't have too much of a strong opinion. Let me know what you think. Otherwise, we could go the other way around and anchor every asset that is ours.

runtime/foucoco/src/xcm_config.rs
gianfra-t: I pushed a new [commit](https://github.com/pendulum-chain/pendulum/pull/432/commits/ddabc241039ce524d50be7912648fba229118a2a) for this. I had to leave `CurrencyId` defined in the common implementation due to some restrictions with the generics, but since all chains share it, I think it should be fine.

runtime/foucoco/src/xcm_config.rs
ebma: I see. IIRC the assets are re-anchored during the processing of XCM messages, so e.g. assuming that chain B sends us our native token with `MultiLocation { parents: 1, parachain: 2094, palletInstance: xx }`, we'll always see it as `MultiLocation { parents: 0, palletInstance: xx }`. I think this re-anchoring will always happen before our custom logic for handling it is called though we probably want to double-check this. Personally, I think it makes most sense if we have all asset multi-locations represented in their anchored form ie. with `parents: 0` if applicable. I checked other parachains and at least centrifuge does this. The other parachains I checked don't seem to have added their own native tokens to their asset registry. 

We can discuss this in a synchronous meeting or invite the other @pendulum-chain/devs to share their opinions here.

runtime/amplitude/src/xcm_config.rs
```diff
@@ -166,11 +91,11 @@ pub type LocalAssetTransactor = MultiCurrencyAdapter<
 	// Use this fungibles implementation
 	Currencies,
 	(), // We don't handle unknown assets.
-	IsNativeConcrete<CurrencyId, CurrencyIdConvert>,
+	IsNativeConcrete<CurrencyId, CurrencyIdConvert<ParachainInfo, AssetRegistryInspector>>,
```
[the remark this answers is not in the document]
gianfra-t: So the naming of the generic `CurrencyIdConvert` is a bit unfortunate, because it should be `AssetRegisterInspector` to be more clear. 

At first I attempted to restrict the second generic argument to an instance of the asset registry itself, but turns out that the methods that we need from the asset registry are not implemented on a specific trait, rather the pallet itself. For instance, [here](https://github.com/pendulum-chain/pendulum/pull/432/files#diff-3287a4ee060b1070a8e6559a094b4890a596a1b281effef54cafe59da20edf49R239) we call the getter for the metadata.

Because of this restriction, I ended up defining this `AssetRegistryInspector` which is just a [wrapper](https://github.com/pendulum-chain/pendulum/pull/432/files#diff-4e8d76261f23472a35cee760b0818583d9267e55b361e44f2c646ab1c031d7d0R64) around the methods we require. I couldn't find a way to make the compiler happy and accept a generic argument that is of type `AssetRegistry`, and it makes sense (I think) since the methods are not on a trait.

runtime/foucoco/src/xcm_config.rs
gianfra-t: I also remember about the re-anchoring, but I am not sure where this happen. I will try to look for it. Also, in centrifuge's they are using the absolute representation and perform the conversion for the `CurrencyIdConvert` type ([here](https://github.com/centrifuge/centrifuge-chain/blob/1a9e385cdc812c4b1a0e55872bbb3576bc48814a/runtime/common/src/xcm.rs#L180))

ebma: Oh right, thanks for linking to that function again. I forgot about the explanation in the doc comment. This indeed suggests that it arrives unanchored. 

Then I suggest we first try to make clear whether this anchoring happens or not (probably done best by running some of our integration tests with logs) and if we see that no re-anchoring happens we leave it as is and change the locations in our asset registry.

runtime/amplitude/src/xcm_config.rs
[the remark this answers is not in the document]
ebma: I see, makes sense to create the extra trait if something like it doesn't exist yet. Thanks for explaining.

runtime/foucoco/src/xcm_config.rs
gianfra-t: Okay so I attempted to use our assets from our chain perspective, yet I hit a roadblock on the following scenario:

Whenever we call `XTokens::transfer` we must specify the asset we want to transfer as `CurrencyId` which the `XTokens` pallet would then [convert](https://github.com/open-web3-stack/open-runtime-module-library/blob/2dbcd1de4db5c05c8596fab447323b4b524673e2/xtokens/src/lib.rs#L409) to a multilocation. Since this now would return a multilocation anchored from our perspective, the execution of such a transfer fails with `Unroutable` error.

This is also probably why previously we [returned](https://github.com/pendulum-chain/pendulum/blob/main/runtime/pendulum/src/xcm_config.rs#L104) the external point of view of our Native token during conversion (thanks @ebma for pointing this out).

So, even though when we receive an asset it is from the native point of view, at least for this scenario described with `XTokens::transfer` we need the asset register to return the external point of view. @TorstenStueber do you think it is okay if we just use the external point of view for the asset registry given this issue?

runtime/common/src/lib.rs
```diff
@@ -242,6 +228,67 @@ pub mod parachains {
...
+/// Here we need to know the canonical representation of all the tokens we handle in order to
+/// correctly convert their `MultiLocation` representation into our internal `CurrencyId` type.
+impl<ParachainId: Get<ParaId>, AssetRegistryInspector: AssetRegistryInspect> xcm_executor::traits::Convert<MultiLocation, CurrencyId> for CurrencyIdConvert<ParachainId, AssetRegistryInspector> {
+	fn convert(location: MultiLocation) -> Result<CurrencyId, MultiLocation> {
+		<CurrencyIdConvert<ParachainId, AssetRegistryInspector> as Convert<MultiLocation, Option<CurrencyId>>>::convert(location)
+			.ok_or(location)
+	}
+}
+
+pub trait AssetRegistryInspect {
```
TorstenStueber, line 287: I don't think that we need to define this trait and then implement it in the `xcm_config.rs` files of the three runtimes. Instead we can just refer to the asset registry of each runtime directly by using the trait 

```
orml_traits::Inspect<
  AssetId = CurrencyId,
  Balance = Balance,
  CustomMetadata = CustomMetadata,
>
```

Then use the trait methods `asset_id` and `metadata`. The asset registry implements the trait [here](https://github.com/open-web3-stack/open-runtime-module-library/blob/4ae0372e2c624e6acc98305564b9d395f70814c0/asset-registry/src/impls.rs#L176-L196).

Then in each runtime you would just use instead of  `CurrencyIdConvert<AssetRegistry>` instead of `CurrencyIdConvert<ParachainId, AssetRegistryInspector>` (see my other comment why `ParachainId` is not required in my opinion).

runtime/amplitude/src/xcm_config.rs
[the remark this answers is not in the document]
TorstenStueber: @gianfra-t I think it is possible, see my other comment.

runtime/foucoco/src/xcm_config.rs
```diff
@@ -240,7 +194,28 @@ impl ShouldExecute for DenyReserveTransferToRelayChain {
...
 
-pub type Barrier = AllowUnpaidExecutionFrom<Everything>;
+match_types! {
+	pub type ParentOrParentsPlurality: impl Contains<MultiLocation> = {
+		MultiLocation { parents: 1, interior: Here } |
+		MultiLocation { parents: 1, interior: X1(Plurality { .. }) }
+	};
+}
+
+pub type Barrier = (
```
TorstenStueber, line 204: This change is not required for this PR. Did you just do it to harmonize Foucoco with the other two runtimes?

TorstenStueber (changes requested):
Overall a good change but I think we should simplify some of the types – I made some proposals in my comments.

runtime/foucoco/src/xcm_config.rs
TorstenStueber: I just read this comments now after I made my review. This explains a little more.

First: let's avoid notions like anchored and unanchored – I don't think they make sense. Every multilocation is relative to something and in the context of our own parachain we should only consider multilocations that are relative to our chain and also use multilocations in the asset registry that are relative to us *without any extra steps one layer up and one layer back down*.

I left some comments to argue that we don't need the `ParachainId` in `CurrencyIdConvert`. Centrifuge might have their own reasons why they do it the way they do it but I don't think that this is a reason for us to copy their behavior, unless we understand their reasoning.

This leaves the problem with `xTokens::transfer`. Can you give a concrete examples that fails? I expect that `xTokens::transfer` works happily with currency ids referenced as (parent: 0, ...) and would assume that we have some misconfiguration here.

runtime/common/src/lib.rs
gianfra-t: Thanks for pointing into the trait!, I thought none was implemented to fetch the metadata. I will remove the unnecessary definition of our trait.

runtime/common/src/lib.rs
```diff
@@ -242,6 +228,67 @@ pub mod parachains {
...
+			.and_then(|m| m.location)
+			.and_then(|l| l.try_into().ok())
+	}
+}
+
+impl<ParachainId: Get<ParaId>, AssetRegistryInspector: AssetRegistryInspect> Convert<MultiLocation, Option<CurrencyId>> for CurrencyIdConvert<ParachainId, AssetRegistryInspector> {
+	fn convert(location: MultiLocation) -> Option<CurrencyId>  {
+		let para_id = ParachainId::get();
+
+		let unanchored_location = match location {
```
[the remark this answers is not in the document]
gianfra-t: The issue I found particularly was with xTokens::transfer. If in our registry we have the location from the local perspective, then the extrinsic will fail since apparently at the level of executing the message, the location must be from the external point of view. I couldn't find a way around this other than having a different currency convert just for this pallet. 

You are correct in the other scenarios that the location (for instance, an incoming transfer) will be anchored from our perspective given that that happens before the message arrives at our chain. 

And about the naming, it means unanchored from our perspective or at least that was the intention. It is essentially the external point of view, that of a sibling parachain.

runtime/foucoco/src/xcm_config.rs
gianfra-t: Yes, that was the intention. Since Foucoco was accepting all messages freely. Of course we can decide to keep it that way. But I think it is good so we are able to test xcm configurations before they reach Amplitude.

runtime/common/src/lib.rs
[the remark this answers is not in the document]
TorstenStueber: I am pretty sure that this is not how it should work. See last section [here](https://github.com/pendulum-chain/pendulum/pull/432#discussion_r1550658905).

Can you give a concrete example? I assume that this is about a reserve transfer deposit from our chain to another chain with an asset originating on our chain. What happens if you do the alternative call `pallet-xcm` → `limitedReserveTransferAssets` instead?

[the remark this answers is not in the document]
gianfra-t: I will try to test that extrinsic in our tests. We can continue with this discussion [here](https://github.com/pendulum-chain/pendulum/pull/432#discussion_r1551818616) since it is about the same and we can have all the information in one place. 
As a side note, I did try `xTokens::transfer_multiasset` which does allow to specify the multi location directly, and I see the same problem: `Unroutable`.

runtime/amplitude/src/xcm_config.rs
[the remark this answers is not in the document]
gianfra-t: I have simplified the code, thanks for the suggestion!

runtime/foucoco/src/xcm_config.rs
TorstenStueber: I had a rough look at how xTokens works and think that the problem occurs [here](https://github.com/open-web3-stack/open-runtime-module-library/blob/4ae0372e2c624e6acc98305564b9d395f70814c0/xtokens/src/lib.rs#L654-L666). Here the XCM message that is executed locally depends on the `transfer_kind`.

Since in this case you intend to move an asset that originates from our chain, that this should behave like a [reserve transfer deposit](https://www.notion.so/satoshipay/23-06-01-Reserve-Transfer-Deposit-910fd65f067c4f2186562494ed7345e8) and should resemble the result of directly calling `pallet-xcm` → `limitedReserveTransferAssets` (I consider xTokens just a set of convenience methods around that).

It becomes clear to me that the behavior we want is the one of the function `transfer_self_reserve_asset` (see https://github.com/open-web3-stack/open-runtime-module-library/blob/4ae0372e2c624e6acc98305564b9d395f70814c0/xtokens/src/lib.rs#L680-L695) because it is the only one of the three functions that has the same result as `pallet-xcm` → `limitedReserveTransferAssets` (see [here](https://www.notion.so/satoshipay/23-06-01-Reserve-Transfer-Deposit-910fd65f067c4f2186562494ed7345e8?pvs=4#c366724ae9684163b67981655fd7d396)).

We get this behavior when the [call to `transfer_kind`](https://github.com/open-web3-stack/open-runtime-module-library/blob/4ae0372e2c624e6acc98305564b9d395f70814c0/xtokens/src/lib.rs#L649C52-L649C71) returns `SelfReserveAsset`.

However, this only happens if `reserve == self_location` [in this line](https://github.com/open-web3-stack/open-runtime-module-library/blob/4ae0372e2c624e6acc98305564b9d395f70814c0/xtokens/src/lib.rs#L825) where:
* `reserve` is the result of [our configuration `ReserveProvider`](https://github.com/pendulum-chain/pendulum/blob/dd8eb3d55ac2933bfc375a13b7a9d2d73ca4ea86/runtime/pendulum/src/xcm_config.rs#L576) applied to the asset multilocation -> [this will just return (parent: 0, Here)](https://github.com/open-web3-stack/open-runtime-module-library/blob/4ae0372e2c624e6acc98305564b9d395f70814c0/traits/src/location.rs#L68-L69)
* `self_location` is the value of our [`SelfLocation` config](https://github.com/pendulum-chain/pendulum/blob/dd8eb3d55ac2933bfc375a13b7a9d2d73ca4ea86/runtime/pendulum/src/xcm_config.rs#L222) and is obviously different

So we need to ensure that `ReserveProvider` for multilocations of the form (parent: 0, ...) return the same as `SelfLocation`.

TorstenStueber: Looking further, we just need to change the SelfLocation to (parent: 0, Here), just see the last but one paragraph [on this page](https://github.com/open-web3-stack/open-runtime-module-library/blob/4ae0372e2c624e6acc98305564b9d395f70814c0/xtokens/README.md):

> Notice the implementation for now also relies on SelfLocation which is already in xtokens config. The SelfLocation can be set to the absolute view (1, Parachain(THIS_PARACHAIN_ID)) and refers to the sender parachain. The alternative is to set SelfLocation to relative view (0, Here) to adhere to Polkadot guidelines.

gianfra-t: Thanks for the investigation, that certainly solves the issue faced by `xTokens::transfer()`, representing our native as `(parent: 0, Here)` introduces some other breaking changes for our current test setup. I will try to explore what exactly fails now. Might have to do with our sibling chain, since now the asset it receives is different.

TorstenStueber: Can you elaborate? Maybe I can help.

gianfra-t: So here I pushed a last [commit](https://github.com/pendulum-chain/pendulum/pull/432/commits/ff45e0c5c552efe8671328e5c0bd5e611af9834e) that implements this idea. I tested both `transfer` and `transfer_multiassets` and they are both working. There were some extra changes to be made though:

- First, in the sibling chain when sending (parents: 0, Here) we now receive the following [multilocation](https://github.com/pendulum-chain/pendulum/pull/432/commits/ff45e0c5c552efe8671328e5c0bd5e611af9834e#diff-84e3c62ba95e693cd5e55dd32c931b4d8ebad427f21b49e2442d288205e24c11R154), which we must handle. I hope this does not break other integrations that other parachains have with us.

- Secondly, and this is only regarding our chain, I had to update a bit the [convert](https://github.com/pendulum-chain/pendulum/pull/432/commits/ff45e0c5c552efe8671328e5c0bd5e611af9834e#diff-3287a4ee060b1070a8e6559a094b4890a596a1b281effef54cafe59da20edf49R247) function. Since `transfer_multiassets` may send already the location from an external point of view, we also need to handle this conversion, that is why we anchor (yes I know the nomenclature is loose but is the name of the method) to our chain all assets before converting. 

- Finally on the same function, we handle the case where the native token recieved is represented as `(0,X1(PalletInstance(10)`, this we cannot control since it comes from other parachains. In this case the testing sibling.

All the changes to the `convert` function are needed since we are now charging the fee on the incoming asset we need to first convert it to access our `asset_metadata`. 

Aside from point 1 which I am a bit afraid may break how other parachains receive our Native, this could work.

node/src/chain_spec.rs
```diff
@@ -537,8 +537,9 @@ fn foucoco_genesis(
 
 	let token_balances = balances
 		.iter()
-		.flat_map(|k| vec![(k.0.clone(), XCM(0), u128::pow(10, 18))])
+		.flat_map(|k| vec![(k.0.clone(), XCM(0), 10*u128::pow(10, 18)), (k.0.clone(), XCM(6), 10*u128::pow(10, 18)), (k.0.clone(), CurrencyId::StellarNative, 10*u128::pow(10, 18))])
```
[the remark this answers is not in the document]
gianfra-t: Probably for testing with Zombienet. I couldn't tell since this PR has undergone many changes. Let me just remove it.

runtime/foucoco/src/xcm_config.rs
TorstenStueber: I'm not sure I follow you. The tests were just wrong in my opinion, it does not correct to – in the context of our parachain – ever refer to an asset as (parent: 1, parachain(id), x) if "id" is our own parachain, instead we need to refer to it as (parent: 0, x).

I don't understand why you would do any reanchoring in `CurrencyIdConvert`. In what context is that required?

gianfra-t: I agree that the test were wrong, in a way. I think it is a matter of choice how to refer to our assets, but now it makes more sense to do it from our perspective. Before, it was consistent with our definition of `SelfLocation` like you were able to find. Now, they agree with the local view.

With respect to the reanchoring, as I said I don't know of anything that would need it right now that is why we can remove it. But since it happened before with pallet `xTokens` which was trying to convert assets represented from an external point of view I imagined this could happen by some other pallet that also uses the convert function.

TorstenStueber: Are there other pallets that use the convert function?

gianfra-t: I will look at that tomorrow and if there are not, then let's remove it!

gianfra-t:
Last commit updated and simplified the implementation now. I will leave a brief description since the conversation is now too long. 

We now don't modify at all the [location](https://github.com/pendulum-chain/pendulum/pull/432/commits/80b5dc580cf61b395632f158d974e4325e98f524#diff-3287a4ee060b1070a8e6559a094b4890a596a1b281effef54cafe59da20edf49R246) before trying to convert nor we convert while [getting](https://github.com/pendulum-chain/pendulum/pull/432/commits/80b5dc580cf61b395632f158d974e4325e98f524#diff-5dc9ab00db16dcf58b342b3f2a9ddd05a114370b9d04b514a3812083382b416dR70) the fee per second.
This is all possible because now we use the local PoV for our assets in the asset registry (example in the [mock](https://github.com/pendulum-chain/pendulum/pull/432/files#diff-dffe390f73891278c96aff2bbe6d1a6146bddc57ce4c8621e0a15aeae822e0c1R236)). `xTokens::tansfer_multiasset` now works with the [local](https://github.com/pendulum-chain/pendulum/pull/432/files#diff-3429c5361e6214779992604d4212c994d6c327c55b74c9bcb280681ad5ddf93dR532) PoV (Thanks @TorstenStueber for figuring that out)

runtime/common/src/asset_registry.rs
gianfra-t: UPDATE: this is now outdated and there is no need to perform this conversion.

runtime/foucoco/src/xcm_config.rs
TorstenStueber: Nice, thanks.

TorstenStueber:
The Foucoco `SelfLocation` is still (parents: 1, Parachain(...)), can you change that please?

TorstenStueber (approved):
This turned out really nice now. Great job!

ebma:
Let's merge and roll it out on Foucoco today?

PR #437, opened 2024-03-20
249 extract chain extension implementation and re use it in all runtimes

### LINKED ISSUE #249: Extract `chain_extension` implementation and re-use it in all runtimes

gianfra-t:
Closes #249 .


chain-extensions/common/src/lib.rs
```diff
@@ -0,0 +1,277 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+#![allow(non_snake_case)]
+
+use scale_info::prelude::vec::Vec;
+use sp_core::{Decode, Encode, MaxEncodedLen};
+use sp_runtime::{codec, ArithmeticError, DispatchError, TokenError};
+
+pub use spacewalk_primitives::{Asset, CurrencyId};
```
ebma, line 8: These are not used here. I guess the idea is to re-export them so they can be used by other modules. However, I saw that the `token` extension also directly imports `spacewalk-primitives` so I'd say we either remove the spacewalk-primitives here or change the import in the `token` extension module.

chain-extensions/price/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+		Err(e) => return Ok(RetVal::Converging(ChainExtensionOutcome::from(e).as_u32())),
+	};
+
+	if let Err(_) = env.write(&result.encode(), false, None) {
+		return Ok(RetVal::Converging(ChainExtensionOutcome::WriteError.as_u32()))
+	};
+	return Ok(RetVal::Converging(ChainExtensionOutcome::Success.as_u32()))
+}
+
+// this was in common, but we don't use it in amplitude's chain extension, is it needed?
```
ebma, line 108: What is this comment referring to?

chain-extensions/price/Cargo.toml
```diff
@@ -0,0 +1,69 @@
...
+runtime-benchmarks = [
+    "frame-system/runtime-benchmarks",
+    "pallet-contracts/try-runtime",
+    "frame-support/runtime-benchmarks",
+]
+
+try-runtime = [
+    "pallet-contracts/try-runtime",
+    "frame-system/try-runtime",
+]
```
ebma, line 68: Should we add these features if we are not intending on using them? Or at least, we can't use them right now as we don't have benchmarks for the extensions.

ebma (commented):
Nice, really great changes! I love it :+1: Just left some minor comments.

chain-extensions/price/src/lib.rs
gianfra-t: Stale comment. Context: at first I was able to use `CoinInfo` directly from `dia` with no problem, and I was planning to remove this re-definition of `CoinInfo`. But I decided to leave it since I was not entirely sure of the implications beyond it compiling.

chain-extensions/price/Cargo.toml
gianfra-t: I was using a template also used for the pallets but we don't need them you are right. It adds more confusion and no benefit. Removing.

ebma (approved):
Looks good to me 👍

TorstenStueber:
Thanks @gianfra-t for this PR. Looks to me like the file `runtime/common/src/chain_ext.rs` can be removed now but it has not been as part of this PR.

gianfra-t:
Yes you are right it should have been removed. I will these changes to the next upcoming PR and reference this one.

PR #440, opened 2024-03-25
Move chain specific definitions

### LINKED ISSUE #383: Move parachain specific definitions out of runtime-common
The `runtime-common` crate contains definitions that are only specific to certain parachains and therefore are better suited as part of the respective parachain crate.

Examples are the `parachains/polkadot` crate, etc.

## TODO
Move these definitions out of the `runtime-common` and into a definition file of the proper runtime crate. In order not to make the `lib.rs` files yet more complex, it is probably best to create a new file, e.g., `parachain.rs`, `constants.rs` or similar and move the definitions there.

gianfra-t:
Closes #383.

Definitions used in only one chain are moved to the new `definitions.rs` file for each runtime. 
Also removes previously hardcoded multi-locations which now are taken from the asset registry. 

Additionally, we remove the common definition of chain extensions which is not needed after #437.


runtime/common/src/lib.rs
```diff
@@ -160,121 +161,27 @@ pub mod parachains {
...
-
-			pub const PARA_ID: u32 = 2040;
-
-			parachain_asset_location!(PDEX, X1(Parachain(PARA_ID)));
-		}
-	}
-
-	pub mod kusama {
+	// These values are shared accross both Pendulum and Kusama
+	pub mod asset_hub {
```
ebma, line 165: I think we have to define this relay-chain specific. If you compare the assets on [Kusama's](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fksm-rpc.stakeworld.io%2Fassethub#/assets) AssetHub to [Polkadot's](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fstatemint.api.onfinality.io%2Fpublic-ws#/assets) Assethub you'll see that the assets are different.

ebma (commented):
I generally like it 👍  we only have to split the asset hub stuff. 

Though I guess that most of the definitions will actually get obsolete once https://github.com/pendulum-chain/pendulum/pull/432 is merged, right? Because we don't necessarily need to keep them around.

gianfra-t:
> I guess that most of the definitions will actually get obsolete once https://github.com/pendulum-chain/pendulum/pull/432 is merged
Yes... I guess it is better to wait for that PR to be merged and check what we actually need, so we avoid cleaning twice.

TorstenStueber:
@gianfra-t would it make change the status of this PR to "Draft" for this reason?

runtime/common/src/lib.rs
```diff
@@ -159,122 +160,16 @@ pub mod parachains {
...
-			parachain_asset_location!(
-				DEV,
-				X2(Parachain(PARA_ID), PalletInstance(BALANCES_PALLET_INDEX))
-			);
-		}
-	}
 }
 
 #[cfg(test)]
 mod tests {
```
b-yap, line 166: these tests do not belong here, I think. `asset_hub` is in amplitude dir.

runtime/amplitude/src/xcm_config.rs
```diff
@@ -28,7 +28,7 @@ use xcm_executor::{
 	XcmExecutor,
 };
 
-use runtime_common::{parachains::kusama::asset_hub, RelativeValue};
+use runtime_common::{parachains::asset_hub, RelativeValue};
```
b-yap, line 31: is it supposed to be `use crate::definitions::asset_hub;` ?

runtime/integration-tests/src/sibling.rs
```diff
@@ -15,7 +15,7 @@ use orml_traits::{
 use pallet_xcm::XcmPassthrough;
 use polkadot_parachain::primitives::Sibling;
 use polkadot_runtime_common::MAXIMUM_BLOCK_WEIGHT;
-use runtime_common::parachains::polkadot::asset_hub;
+use runtime_common::parachains::asset_hub;
```
b-yap, line 18: also need to update this.

runtime/amplitude/src/xcm_config.rs
gianfra-t: Yes! Thanks. We later moved the `asset_hub` definition to each chain.

gianfra-t:
After the upgrade from #432 to use the asset registry, I believe we will no longer need many "hardcoded" locations defined before (for instance, [these](https://github.com/pendulum-chain/pendulum/pull/440/files#diff-dc2adfc16e0b81b01c0a583075789c42335dfd48c3802819037b1929d7b494f5L10))

runtime/amplitude/src/definitions.rs
```diff
@@ -0,0 +1,37 @@
...
+#[cfg(test)]
+mod tests {
+	use super::asset_hub;
+	use xcm::{
+		latest::prelude::{PalletInstance, Parachain},
+		prelude::GeneralIndex,
+	};
+
+	#[test]
+	fn test_pink() {
```
bogdanS98, line 22: Also this should be moved to Pendulum definitions

bogdanS98 (changes requested):
Looks good to me 👍🏼 Though we should move PINK definitions to Pendulum.

gianfra-t:
@pendulum-chain/devs I have removed even more definitions that I believe we only used for currency conversion in the past. See for instance, [here](https://github.com/pendulum-chain/pendulum/pull/440/commits/35b497a24277a026ef99a9ed54d9eb0368c9e97b#diff-ea7fa78385c128b4b9005015219ec0b2c03a2a076664ade4fccc9a5436331545L8). The runtime compiles but I am afraid I am missing some use case for this constants, so please check.

I decided to leave `definitions.rs` even though is empty for some chains in case we want to add some in the future.

runtime/foucoco/src/assets.rs
```diff
@@ -5,34 +5,3 @@ pub mod xcm_assets {
 	create_xcm_id!(RELAY, 0);
```
TorstenStueber, line 5: These definitions are not necessarily part of this PR but I don't think that they are right.

In the asset registry XCM(1) on Foucoco refers to "rTether USD" on the Rococo Asset Hub. XCM(2) refers to the native asset of some parachain with para id (2087) – I don't know which one that it, it is not online.

runtime/common/src/chain_ext.rs
```diff
@@ -1,300 +0,0 @@
-use crate::*;
```
TorstenStueber, line 1: Why is this unrelated file removed here?

TorstenStueber (commented):
I am not sure whether it makes sense to have an `assets.rs` and a `definition.rs` file per runtime. At the end the definitions inside `definition.rs` are all asset related and for that reason it would make sense to just join both files.

On the other hand: are the remaining definitions in `assets.rs` still needed? At the end we don't want to have a need to execute runtime upgrades when we register new XCM assets – that's why we moved all the logic into the asset registry pallet. But if we register new assets without changing the code, then the definitions in `asset.rs` deviate from the truth.

Looks to me like these definitions are only needed in the integration tests, correct? Is there a way to load the information from the asset-registry storage inside the tests instead and then we can remove the remaining content of all `asset.rs` files.

runtime/common/src/chain_ext.rs
bogdanS98: I think it's because of [this](https://github.com/pendulum-chain/pendulum/pull/437) already merged PR

TorstenStueber: Ah, yes, sure. Because of [this comment](https://github.com/pendulum-chain/pendulum/pull/437#issuecomment-2031746174).

runtime/foucoco/src/assets.rs
gianfra-t: I removed them since we also don't seem to be using this xcm assets definitions.

gianfra-t:
@TorstenStueber there are very few definitions from the `assets` files actually used, which I now moved to the pendulum definitions file. You are correct that they are only used in our integration tests (except for BRZ which is also used for the automation [routing](https://github.com/pendulum-chain/pendulum/blob/cd2a70799f0ffd3d516f91a310e3a435081d2203/runtime/pendulum/src/xcm_config.rs#L214) for the BRZ token).

We use these definitions [here](https://github.com/pendulum-chain/pendulum/blob/main/runtime/integration-tests/src/sibling.rs#L159) for defining the currency conversion of the sibling chain but also in our testing [configuration](https://github.com/pendulum-chain/pendulum/blob/cd2a70799f0ffd3d516f91a310e3a435081d2203/runtime/integration-tests/src/mock.rs#L258) of the asset registry for pendulum and amplitude. For this reason I think these definitions are useful even when we use the asset registry because it saves us having to manually define the location.

Alternatively they could be moved to the integration test directory (except for BRZ).

TorstenStueber:
I prefer to move definitions that are only used for testing into testing related code in order to not "clutter up" the rest. What do you think?

gianfra-t:
Yes! It makes sense. Moved to a `definitions.rs` in the `integration-tests` folder.

TorstenStueber (approved):
Nice, thanks!

bogdanS98 (approved):
Looks good 👍🏼

PR #453, opened 2024-04-16
178 add benchmarks for custom pallets per runtime

bogdanS98:
Closes #178.

PR #459, opened 2024-05-03
384 add clippy to ci

### LINKED ISSUE #384: Add clippy to CI
It would probably be beneficial to add clippy checks, as they are [disabled](https://github.com/pendulum-chain/pendulum/blob/58d806850b912e22905c444d99db7a9b0757df98/.github/workflows/test-code.yml#L48-L50) right now. We can use the same [rules](https://github.com/pendulum-chain/spacewalk/blob/a5cc81ececb9a464fc8caa12c191eeaecf453c4b/.github/workflows/ci-main.yml#L80-L91) we already use in Spacewalk.

gianfra-t:
Closes #384  

Implements the same two rules for checking the code with `clippy` [used](https://github.com/pendulum-chain/spacewalk/blob/main/.github/workflows/ci-main.yml#L78) in Spacewalk, one more strict for the main targets, and one more forgiving for the remaining ones.

Due to the different imports in the package `runtime-integration-tests`, we need to run a separate checking [step](https://github.com/pendulum-chain/pendulum/pull/459/files#diff-9d8a210f3bc0ae57dbd26c81f07a2811ad8b593f2b125e5b37e37522f90ff644R85) only for this package for it to compile properly. The rules are the same as for all targets.

Additionally, some modifications are made to avoid some errors and warnings now that the check is implemented.


ebma (commented):
Looks good to me 👍 Let's try to fix all the warnings that are now shown in this PR though.

gianfra-t:
Yes there are many warnings about unused functions that I believe happen only because of how `clippy` is analyzing the crates in an isolated manner. I will look for a flag to skip that warnings. 

On the other hand, there are some warnings (that I rolled back before) because I thought it was modifying a very critical part of the chain extension in this case which is hard to test properly. I will push that again and we can discuss it.

chain-extensions/token/src/lib.rs
```diff
@@ -219,7 +218,7 @@ where
 		"Calling transfer() sending {:?} {:?}, from {:?} to {:?}",
 		amount,
 		currency_id,
-		env.ext().caller().clone(),
+		&(*env.ext().caller()).clone(),
```
ebma, line 221: About all these 'clone-double-ref' changes, I looked into what exactly this means and I guess the easiest change would be to just use `env.ext().caller()` each time, without any cloning, since the `.caller()` function already returns a reference. @gianfra-t Can you please try if changing all the `&(*env.ext().caller()).clone()` to just `env.ext().caller()` works? Maybe I'm missing something.

gianfra-t: Wow yes I missed that. I'm surprised clippy even suggested this vs just removing clone. But you are right it returns a reference to account id already, and all methods require a reference only.

ebma (approved):
Looks good to me, I think we can merge once the CI passed 👍

gianfra-t:
We can merge this after fixes that will come shortly [here](https://github.com/pendulum-chain/pendulum/pull/453) for compiling with `--all-features`.

runtime/amplitude/src/lib.rs
```diff
@@ -85,11 +85,13 @@ pub use sp_runtime::BuildStorage;
 pub use dia_oracle::dia::AssetId;
 pub use issue::{Event as IssueEvent, IssueRequest};
 pub use nomination::Event as NominationEvent;
+#[allow(unused_imports)]
```
TorstenStueber, line 88: Why do we need to keep this import?

runtime/integration-tests/src/pendulum_tests.rs
```diff
@@ -13,16 +13,20 @@ use crate::{
 };
 
 use frame_support::assert_ok;
+#[allow(unused_imports)]
 use pendulum_runtime::definitions::moonbeam::PARA_ID as MOONBEAM_PARA_ID;
```
TorstenStueber, line 17: Do we need to keep this import?

TorstenStueber (approved):

runtime/amplitude/src/lib.rs
gianfra-t: I separated and added the tag to remove the unused warning when we compile with with the `runtime-benchmarks`, because it is not [used](https://github.com/pendulum-chain/pendulum/blob/12d19aa4ead6aaa8a42530b3b47f0787d90d5b78/runtime/amplitude/src/lib.rs#L275) in that case.

bogdanS98 (approved):
Looks good to me too! 👍🏼 @gianfra-t [#178](https://github.com/pendulum-chain/pendulum/pull/453) is merged now FYI.

gianfra-t:
@pendulum-chain/devs I was not able to remove the warning for `impl X::Config for Runtime {}`, I don't think it's critical to remove it, so I will just merge this and later we can explore how to refactor so it doesn't warn.

ebma:
Can we just disable the 'non_local_definitions' rule for the clippy checks then? I remember I was also struggling with this in Spacewalk, IIRC @b-yap and I just changed back the Rust toolchain version to an older one but that's of course not ideal. 
Also, let's change the CI so that warnings get denied with [this](https://github.com/pendulum-chain/spacewalk/blob/53f81a2dac91d523eaf7178d01a402a0fa536d55/.github/workflows/ci-main.yml#L17) flag. Otherwise the clippy checks are easily missed/ignored. If we deny every warning and ignore the 'non_local_definitions' error, we need to fix the three other minor issues shown (use of constant weight and unused imports).

gianfra-t:
@ebma thanks for the linting reference, I thought about ignoring the local `impl` but couldn't find the lint name to `allow`. 

I also replaced the use of the direct definition of the weight of pallet vesting [here](https://github.com/pendulum-chain/pendulum/pull/459/commits/a765f14d174bb90a3f712dd5cef14689db58297b#diff-12a0252d4a88f4179d99aa18c3b07ba66635a90f9bff12ddb6c3f6fc0875f97bR35). It is still manual, but at least clippy will not complain. Otherwise we need to do a benchmark just for this extrinsic.

ebma (approved):
Great 👍 looks good to me

PR #462, opened 2024-05-10
Use Foucoco runtime with instant seal

gianfra-t:
Notes:
This does involve changes to the node of course, but it is not required for collators to update since it involves modifications for testing the chain locally.

Closes [tasks/#235](https://github.com/pendulum-chain/tasks/issues/235).

### DIFF
--- a/node/src/command.rs
+++ b/node/src/command.rs
@@ -8,10 +8,9 @@ use log::{info, trace};
 use runtime_common::opaque::Block;
 use sc_chain_spec::GenericChainSpec;
 use sc_cli::{
-	ChainSpec, CliConfiguration, DefaultConfigurationValues, ImportParams, KeystoreParams,
-	NetworkParams, Result, RuntimeVersion, SharedParams, SubstrateCli,
+	ChainSpec, CliConfiguration, DefaultConfigurationValues, ImportParams, KeystoreParams, NetworkParams, Result, RuntimeVersion, SharedParams, SubstrateCli
 };
-use sc_service::config::{BasePath, PrometheusConfig};
+use sc_service::{Configuration, config::{BasePath, PrometheusConfig}};
 use sp_core::hexdisplay::HexDisplay;
 use sp_runtime::traits::{AccountIdConversion, Block as BlockT};
 
@@ -194,22 +193,24 @@ macro_rules! construct_sync_run {
 	(|$components:ident, $cli:ident, $cmd:ident, $config:ident| $code:expr ) => {{
 		let runner = $cli.create_runner($cmd)?;
 
+		// none of these are using instant seal
+		let is_standalone = false;
 		match runner.config().chain_spec.identify() {
 			ChainIdentity::Amplitude => runner.sync_run(|$config| {
 				let $components = new_partial::<
 					amplitude_runtime::RuntimeApi,
 					AmplitudeRuntimeExecutor,
-				>(&$config)?;
+				>(&$config, is_standalone)?;
 				$code
 			}),
 			ChainIdentity::Foucoco => runner.sync_run(|$config| {
 				let $components =
-					new_partial::<foucoco_runtime::RuntimeApi, FoucocoRuntimeExecutor>(&$config)?;
+					new_partial::<foucoco_runtime::RuntimeApi, FoucocoRuntimeExecutor>(&$config, is_standalone)?;
 				$code
 			}),
 			ChainIdentity::Pendulum => runner.sync_run(|$config| {
 				let $components =
-					new_partial::<pendulum_runtime::RuntimeApi, PendulumRuntimeExecutor>(&$config)?;
+					new_partial::<pendulum_runtime::RuntimeApi, PendulumRuntimeExecutor>(&$config, is_standalone)?;
 				$code
 			}),
 		}
@@ -220,22 +221,24 @@ macro_rules! construct_generic_async_run {
 	(|$components:ident, $cli:ident, $cmd:ident, $config:ident| $code:expr ) => {{
 		let runner = $cli.create_runner($cmd)?;
 
+		// none of these are using instant seal
+		let is_standalone = false;
 		match runner.config().chain_spec.identify() {
 			ChainIdentity::Amplitude => runner.async_run(|$config| {
 				let $components = new_partial::<
 					amplitude_runtime::RuntimeApi,
 					AmplitudeRuntimeExecutor,
-				>(&$config)?;
+				>(&$config, is_standalone)?;
 				$code
 			}),
 			ChainIdentity::Foucoco => runner.async_run(|$config| {
 				let $components =
-					new_partial::<foucoco_runtime::RuntimeApi, FoucocoRuntimeExecutor>(&$config)?;
+					new_partial::<foucoco_runtime::RuntimeApi, FoucocoRuntimeExecutor>(&$config, is_standalone)?;
 				$code
 			}),
 			ChainIdentity::Pendulum => runner.async_run(|$config| {
 				let $components =
-					new_partial::<pendulum_runtime::RuntimeApi, PendulumRuntimeExecutor>(&$config)?;
+					new_partial::<pendulum_runtime::RuntimeApi, PendulumRuntimeExecutor>(&$config, is_standalone)?;
 				$code
 			}),
 		}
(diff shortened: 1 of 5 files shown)


TorstenStueber:
There are a couple of changes in this PR that are pure formatting changes in unrelated code. How come, is the code on the main branch currently not correctly formatted or is did you reformat the code with different parameters? Makes it harder to review this PR.

TorstenStueber:
It seems that actually the current code on main is not correctly formatted. For that reason [I added this issue](https://github.com/pendulum-chain/pendulum/issues/477).

Is it possible to revert the formatting changes in unrelated code to keep this PR focussed and then correct the formatting when working on the linked issue?

node/src/chain_spec.rs
```diff
@@ -169,6 +192,101 @@ pub fn foucoco_config() -> FoucocoChainSpec {
...
+	];
+
+	signatories.extend(prefunded_accounts.iter().cloned());
+
+	let invulnerables: Vec<_> = foucoco::INITIAL_COLLATORS
+		.iter()
+		.map(|ss58| AccountId::from_ss58check(ss58).unwrap())
+		.collect();
+
+	let sudo_account = get_account_id_from_seed::<sr25519::Public>("AltoParaíso"); //pallet_multisig::Pallet::<Runtime>::multi_account_id(&signatories[..], 3);
```
TorstenStueber, line 255: We can of course use a different seed here. I just happened to be in the town of Alto Paraíso when I wrote this for foucoco-standalone and wasm-deploy.

node/src/command.rs
```diff
@@ -394,106 +422,129 @@ pub fn run() -> Result<()> {
...
-				let genesis_state = format!("0x{:?}", HexDisplay::from(&block.header().encode()));
-
-				let tokio_handle = config.tokio_handle.clone();
-				let polkadot_config =
-					SubstrateCli::create_configuration(&polkadot_cli, &polkadot_cli, tokio_handle)
-						.map_err(|err| format!("Relay chain argument error: {err}"))?;
+			runner
+				.run_node_until_exit(|config| async move {
+					if cli.instant_seal {
+						start_instant(config).await
```
TorstenStueber, line 437: I don't understand what all the code in `start_node` is doing. Sure that we need none of this for the foucoco-standalone version?

TorstenStueber (commented):
I can't evaluate the code in `node/service.rs`, it's just code I don't understand. Was this just inspired by taking the code from foucoco-standalone? Is it possible to unify this further? I can imagine that it becomes a major pain point when updating Polkadot versions.

node/src/command.rs
```diff
@@ -194,24 +203,33 @@ macro_rules! construct_sync_run {
...
-					new_partial::<pendulum_runtime::RuntimeApi, PendulumRuntimeExecutor>(&$config)?;
+				let $components = new_partial::<
+					pendulum_runtime::RuntimeApi,
+					PendulumRuntimeExecutor,
+				>(&$config, is_standalone)?;
 				$code
 			}),
+			// Foucoco standalone is only supported
+			// with the instant seal flag.
+			ChainIdentity::FoucocoStandalone => todo!(),
```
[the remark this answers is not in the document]
TorstenStueber: In the latter case, I think we should rather mark it as `unreachable!` if logically impossible or as `unimplemented!` if we don't actually intend to implement that case.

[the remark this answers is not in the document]
gianfra-t: Right, I forgot about those macros. I believe `unimplemented` makes more sense since it can actually be called with those flags, they just don't make sense for now.

node/src/chain_spec.rs
gianfra-t: Does Alice sound good? It is easier to keep track to.

node/src/command.rs
gianfra-t: As far as I am aware it is mostly to start the relay configurations, which we don't use (and in fact we don't want) in the standalone [version](https://github.com/pendulum-chain/foucoco-standalone/blob/main/node/src/command.rs#L122).

gianfra-t:
> It would be nice to have this how-to at the bottom of the README.md. 
@b-yap Totally! I was waiting to see if in review something came up and we decided to change the commands or implementation, as soon as we define it I will add this to the README.md. 

@TorstenStueber Regarding formatting, I realized after the fact that it was a bad idea for reviewing the changes, so I rolled back the format and will maybe do at the end before merging. 
Regarding the changes in `node/service.rs`, I took inspiration mostly from how we start the `foucoco-standalone` service. 

The idea was to maintain the same configurations of services we have right now for the production chain, as well as the same standalone uses without duplicating all the code. I tried to unify as much as I could the common initialization between both [here](https://github.com/pendulum-chain/pendulum/pull/462/files#diff-594440db8adfc5d1b37711309312bfa06b49e488309186ad062a896778cf9fbfR292), but it became cumbersome due mostly to `parachain_config`[here](https://github.com/pendulum-chain/pendulum/pull/462/files#diff-594440db8adfc5d1b37711309312bfa06b49e488309186ad062a896778cf9fbfR397) which does not implement clone which made it hard to move it to a common function implementation. And yes this all will probably change substantially when we upgrade.

node/src/chain_spec.rs
TorstenStueber: I think it is straightforward to use Alice but I think there was a reason why I did not want to use it. I just don't remember ... let's just use Alice then.

node/src/command.rs
TorstenStueber: To better understand this: is the new foucoco-standalone runtime you are defining here a standalone substrate chain or a parachain (i.e., it also requires a relay chain?)

ebma (commented):
I'm not a big fan of the standalone naming but I don't necessarily have a better word. 
How about:
- `foucoco-testing`
- `foucoco-instant`
- or just `dev`. We previously had a `dev` runtime but removed it as we never used it. We can reintroduce the name and call it `dev` while pointing to the foucoco runtime with a different chain spec.

node/src/command.rs
gianfra-t: It does not require a relay chain and works like the standalone version in the repo (foucoco-standalone repo).

TorstenStueber: But then wouldn't we need to define a completely different runtime with a different set of pallets? Don't the cumulus pallets require a connection to a relay chain, etc?

gianfra-t: At least during testing I was able to run the deployment of the nabla contracts, I believe it ignores the runtime logic that relates to the real consensus mechanisms (cumulus pallets, for example). I thought @ebma was referring just to the name(?), but this is an important point. 

Do you think testing/deploying `nabla` (or any `wasm-deploy` script, etc) is not enough to ensure that the runtime logic works on this standalone version, even with those pallets?

TorstenStueber: If the standard foucoco runtime (that is intended to be a parachain) really happens to work as a standalone chain, then that's awesome and should be enough for the Nabla tests. I seriously did not expect that to work. Did you try to run the Nabla test suite with it?

ebma: Interesting, I honestly assumed that the new instant seal runtime would run like the others. If it's running like a standalone chain, there is no need to rename it after all. Seems like the best of both worlds for us, except the overhead of maintaining the extra logic in the node but we probably cannot avoid that. 

We would need to ask ourselves if it makes any difference to our testing setup. In the past, we noticed that the block weight limits were different between the standalone chain and our parachains. I didn't see any changes related to this in the code so I would now naively assume that we don't face this issue going forward as the configuration is shared between standalone and parachain runtime. Anything else that comes to mind that we should pay attention to @gianfra-t or @TorstenStueber?

ebma (commented):
I tested it locally and can confirm that it works. I ran the chain with 
`cargo run -p pendulum-node -- --chain foucoco-standalone --instant-seal` and was able to do a simple transfer that got executed instantly.

@gianfra-t let's add some details how to run the chain with instant-seal to the README then.

node/src/command.rs
gianfra-t: Yes it was a pleasant surprise that the runtime worked without modifications. When I ran the Nabla test suite I encountered the known "contract too long" error `CodeTooLarge`, so I couldn't verify completely.

But deploying also calls the chain extensions in order to approve so I have reason to believe all the components work. The only problem I've encountered is because the db gets corrupted or conflicts with a previous run, I don't know how this was handled before but to solve I either delete it or change the id of the chain temporarily (which creates a new db). 

@ebma am not sure about weight limits, but the configurations must now the same. It is hard to think of another issue that could arise.

node/src/chain_spec.rs
```diff
@@ -169,6 +192,101 @@ pub fn foucoco_config() -> FoucocoChainSpec {
...
+		Some(properties),
+		// Extensions
+		ParachainExtensions {
+			relay_chain: "kusama".into(), // You MUST set this to the correct network!
+			para_id: foucoco::PARACHAIN_ID,
+		},
+	)
+}
+
+pub fn foucoco_standalone_config() -> FoucocoChainSpec {
```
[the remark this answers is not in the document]
gianfra-t: You think a different name will be better after? Since we where talking about the naming also [here](https://github.com/pendulum-chain/pendulum/pull/462#discussion_r1611445952).

node/src/service.rs
```diff
@@ -390,25 +466,25 @@ where
 
 	if validator {
 		let parachain_consensus = build_consensus(
-			client.clone(),
+			client_clone.clone(),
```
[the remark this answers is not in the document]
gianfra-t: Good observation, there is no reason. Let's clone the client directly.

node/src/chain_spec.rs
gianfra-t: I changed to Alice, but we should remember also to change in `wasm-deploy` locally.

node/src/chain_spec.rs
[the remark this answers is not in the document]
ebma: No no, nevermind. It's like the standalone chain after all.

node/src/command.rs
ebma: >  The only problem I've encountered is because the db gets corrupted or conflicts with a previous run, I don't know how this was handled before but to solve I either delete it or change the id of the chain temporarily (which creates a new db).

How exactly did you encounter this issue?

> I  am not sure about weight limits, but the configurations must now the same

Yeah I guess we can assume they are the same 👍

node/src/service.rs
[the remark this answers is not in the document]
ebma: In the same file there are more usages of `client_clone.clone()`, let's change them too 😬

ebma:
The README looks good, thanks @gianfra-t 👍  seems like [something went wrong](https://github.com/pendulum-chain/pendulum/pull/462/commits/34ed630c4856160ed9adcb693fd59546e39fda98#diff-3287a4ee060b1070a8e6559a094b4890a596a1b281effef54cafe59da20edf49R201) in your merge commit, that's why the CI is complaining.

gianfra-t:
Ohh yes I was hoping I caught all conflicts. My vs code was buggy and not showing them.

gianfra-t:
@ebma I think the new build [error](https://github.com/pendulum-chain/pendulum/actions/runs/9210982964/job/25339255586?pr=462#step:8:822) has to do with the runner changes, any idea what can cause it ?

ebma:
Yes, @b-yap mentioned a solution [here](https://github.com/pendulum-chain/spacewalk/pull/519#discussion_r1609826532). It should work if you make sure that the `ahash` dependency used is at least v0.8.x I think.

gianfra-t:
Thanks! `cargo update` did the trick. `0.8.3` was the problematic one.

node/src/command.rs
TorstenStueber: @ebma It could be hard to research whether there are some subtle differences. For now at least it might be enough if we can use it for Nabla (or similar smart contracts). Do you see other uses for the standalone version?

The "contract too long" error is a problem to run the test suit but it should be unrelated to foucoco-standalone.

I don't know how you mean that the db gets corrupted. When I run the test suite, I always start everything from scratch: foucoco-standalone starting at genesis and the indexer running locally with an empty db.

gianfra-t: About the db corruption: Since while testing I started the node many times, I found two strange errors: the node crashing on star and the deployment of the contracts not being successful due to failure in a chain extension call. 

For the first case, it happened that at the beginning when I was using the same config name `foucoco` for the standalone version, this meant that upon start the node would use the already existing db used for `foucoco`. Now, I don't know why exactly this happened, but simply changing the [name](https://github.com/pendulum-chain/pendulum/pull/462/files/1607e16b522fbe71946be7e67dd063a0d484616e#diff-f2256017c4e606a30f53708578da46675512cc1300be6542ed5f77e58af0d6dcR264) to, for example, `foucoco2` would solve the issue. This happened only these two times and I couldn't replicate. I suspect it was the db since it was the only variable.

node/src/service.rs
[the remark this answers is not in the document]
gianfra-t: Oh...you are right, I got lost with so many changes. That extra clone was needed for a closure that I modified later. [Here](https://github.com/pendulum-chain/pendulum/pull/462/commits/e1bf7e683734aae77a24b49ce10ef9ae1f97a202) I removed the second clone and return directly the first one at the end. We do need `client_clone_move` though.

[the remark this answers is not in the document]
ebma: I see 👍

ebma (approved):

PR #463, opened 2024-05-13
#444 Switch to 'stable' rust toolchain channel

### LINKED ISSUE #444: Switch to 'stable' rust toolchain channel
At the moment, we are using the 'nightly' channel of the Rust toolchain. Going forward, we want to use the 'stable' channel because some tools like the SRTool don't work on 'nightly' anymore, at least not on the latest versions, see [this](https://github.com/paritytech/srtool/issues/62) issue. 

Switching the toolchain was not possible before as we encountered compiler errors. It might be that the recent update of the polkadot dependencies to v0.9.42 resolves those problems.

## TODO
Change the `rust-toolchain.toml` to use a 'stable' channel. Make sure that all runtimes and integration tests still compile and work as expected.

b-yap:
Closes #444. 

There is a problem with using the Rust `nightly-2024-02-09`:
```
error[E0449]: visibility qualifiers are not permitted here
   --> /Users/b.carlayap/.cargo/git/checkouts/substrate-7e08433d4c370a21/ff24c60/frame/contracts/src/lib.rs:975:3
    |
975 |         environmental!(executing_contract: bool);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: trait items always share the visibility of their trait
    = note: this error originates in the macro `environmental` (in Nightly builds, run with -Z macro-backtrace for more info)
```
I tried a lower version (`nightly-2023-12-28`) but errors still abound:
```
error[E0658]: use of unstable library feature 'stdsimd'
  --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/curve25519-dalek-4.1.2/src/backend/vector/ifma/field.rs:26:5
   |
26 |     _mm256_madd52lo_epu64(z.into(), x.into(), y.into()).into()
   |     ^^^^^^^^^^^^^^^^^^^^^
   |
   = note: see issue #48556 <https://github.com/rust-lang/rust/issues/48556> for more information
   = help: add `#![feature(stdsimd)]` to the crate attributes to enable
```
Found [this comment](https://github.com/succinctlabs/sp1/issues/500#issuecomment-2063144675) to choose a newer version, so I picked `nightly-2024-04-18`.

----  
## How to begin the review:   
1. .github/actions/shared/action.yml:
   * we have to separate the jobs for building project using the stable version, and running the tests with nightly. But both jobs have similar steps. The shared steps are transferred in this file.
   * the file name `action.yml` is a must. See [documentation](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#about-yaml-syntax-for-github-actions):
      > The metadata filename must be either action.yml or action.yaml.
2. .github/workflows/test-code.yml:
    * split into 2 jobs:
       *  build-check - to run `cargo check` with stable version
       * test-code - to run `cargo test` with nightly version
3. README.md:
   * add explanation on why we override the toolchain for testing
4. rust-toolchain.toml:
   * change to stable `1.77.0`
5. the other files are to clear up build warnings.


.github/workflows/test-code.yml
```diff
@@ -70,5 +32,35 @@ jobs:
...
+          components: rustfmt, clippy
+          target: wasm32-unknown-unknown
+
+      - name: Setup nightly Rust as default
+        run: rustup default nightly-2024-04-18
+
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v1
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
b-yap, line 60: somehow, accessing the secrets in the action file will cause an error. I had to call it here, on both jobs: **build-check** and **test-code**.

ebma (approved):
Looks good to me 👍 A bit annoying that here we again have to do some extra handling for nightly due to the mocktopus crate but doing it this way is probably easier than refactoring the code to remove the mocktopus crate entirely. And in the future we might encounter other crates that we want to have in tests but need nightly so IMO it's okay to have this stable vs nightly setup you added @b-yap.

PR #468, opened 2024-05-15
Add EURC to Zenlink ID map.

gianfra-t:
Closes: [tasks/317](https://github.com/pendulum-chain/tasks/issues/317).
Related: [Pair creation](https://github.com/pendulum-chain/transaction-builder/pull/109).

In order to add a Zenlink pair with stellar asset `EURC`, we need to add it manually to both mapping functions for Zenlink.


ebma (approved):
Nice catch, I already forgot that we have this mapping 👍

vadaynujra:
So this step would be required each time we want to add any of the Spacewalk assets to a pool on Zenlink @gianfra-t ?

ebma:
Yes, unfortunately. We can already add potential support for other Stellar assets in advance but they have to be added to this list. So if you already have some Spacewalk assets in mind that you want to have in the near future, let us know.

vadaynujra:
I think it makes sense to keep that option, without having to do a runtime upgrade each time. Also, I don't see any downside of already mapping them. Therefore, updated the description of the ticket to include all Spacewalk assets.

gianfra-t:
@vadaynujra since we already close this PR I am afraid we will miss this. I can add it as a to-do list in the [original](https://github.com/pendulum-chain/tasks/issues/317) task if you like.

vadaynujra:
@gianfra-t would that mean we add the mapping functions for the assets also as part of the next runtime upgrade? If yes, please do that.

ebma:
If it's very urgent I can also quickly add them now before preparing the runtime upgrade for tomorrow. Otherwise you'd have to wait probably a month or so, not sure when we do the next runtime upgrade on Pendulum.

PR #470, opened 2024-05-16
Replace oracle key convertor

### LINKED ISSUE #469: Use asset registry metadata for implementation of `DiaOracleKeyConvertor`
## Context

Spacewalk needs to map the specific types of our `CurrencyId` enum to the respective asset prices stored in the DIA pallet. These prices are identified by a 'blockchain' + 'symbol' pair. Currently, we are using hard-coded mappings, see eg [here](https://github.com/pendulum-chain/pendulum/blob/cafe9df38ff623ba074e0efa9d856b69f01da999/runtime/foucoco/src/lib.rs#L199C12-L225) but we want to use the metadata of the asset registry instead so that we don't need to touch that mapping again.

## TODO

Replace the hard-coded mapping with logic that checks the metadata of the asset registry.

ebma:
Introduces a new struct that converts between `CurrencyId` <> `DiaKeys` using the metadata in the asset registry. The lookup of `DiaKeys -> CurrencyId` can be costly because we are potentially iterating over all elements of the `StorageMap` but since we don't have that many assets in the registry it should be fine.

Closes #469.

### DIFF
--- a/runtime/common/src/asset_registry.rs
+++ b/runtime/common/src/asset_registry.rs
@@ -108,6 +108,10 @@ impl<
 	> Convert<(Vec<u8>, Vec<u8>), Option<Key>> for AssetRegistryToDiaOracleKeyConvertor<Runtime>
 {
 	fn convert(dia_oracle_key: (Vec<u8>, Vec<u8>)) -> Option<Key> {
+		if dia_oracle_key.0.is_empty() || dia_oracle_key.1.is_empty() {
+			return None;
+		}
+
 		// Try to find the currency id in the asset registry metadata for which the dia keys are
 		// matching the ones provided
 		let blockchain = dia_oracle_key.0;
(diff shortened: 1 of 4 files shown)


runtime/common/src/asset_registry.rs
```diff
@@ -70,4 +74,50 @@ impl<
...
+		// We didn't find the dia keys in the asset registry metadata
+		None
+	}
+}
+
+impl<
+		Runtime: orml_asset_registry::Config<AssetId = CurrencyId, CustomMetadata = CustomMetadata>,
+	> Convert<(Vec<u8>, Vec<u8>), Option<Key>> for AssetRegistryToDiaOracleKeyConvertor<Runtime>
+{
+	fn convert(dia_oracle_key: (Vec<u8>, Vec<u8>)) -> Option<Key> {
```
bogdanS98, line 110: Is it possible to have tuples that contain empty `Vec<u8>` here? In that case, should we check for empty vectors so we spare some storage iterations?

bogdanS98 (commented):
Overall looks good to me, only left one question 👍🏼

runtime/common/src/asset_registry.rs
ebma: Good point. I added an additional check, just in case.

bogdanS98 (approved):

PR #471, opened 2024-05-16
445 replace oraclekey enums used in treasury buyout extension

### LINKED ISSUE #445: Replace OracleKey enums used in treasury-buyout-extension
### Context

At this moment, `treasury-buyout-extension` is using `oracle` pallet from Spacewalk to fetch asset prices. Since `CustomMetadata` struct containing DIA keys was added to our runtimes, we can use `dia-oracle` pallet directly for this use case.

### Requirement

Modify `PriceGetter` implementation for all runtimes so that it uses DIA fields from `CustomMetadata` struct instead of `OracleKey` [enum](https://github.com/pendulum-chain/pendulum/blob/dd8eb3d55ac2933bfc375a13b7a9d2d73ca4ea86/runtime/pendulum/src/lib.rs#L1409-L1410).

Note: DIA fields should be checked if they are empty.

bogdanS98:
This PR modifies `PriceGetter` trait implementation by replacing usage of `oracle` pallet with `dia-oracle`. 
`PriceGetter` is a trait that is used to fetch prices of assets when doing buyouts.

Closes #445.

### DIFF
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -3,13 +3,13 @@
 
 use asset_registry::CustomMetadata;
 use core::{fmt::Debug, marker::PhantomData};
-use dia_oracle::{DiaOracle, CoinInfo};
+use dia_oracle::{CoinInfo, DiaOracle};
 use orml_traits::asset_registry::Inspect;
-use sp_std::vec;
 use sp_runtime::{
 	traits::{Convert, IdentifyAccount, One, Verify, Zero},
 	DispatchError, FixedPointNumber, FixedU128, MultiSignature,
 };
+use sp_std::vec;
 use spacewalk_primitives::CurrencyId;
 use treasury_buyout_extension::PriceGetter;
 use xcm::v3::{AssetId, MultiAsset, MultiLocation};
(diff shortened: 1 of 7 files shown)


ebma (commented):
Let's generify and deduplicate the `OraclePriceGetter` and put it into the `common` directory, similar to what I did with the struct [here](https://github.com/pendulum-chain/pendulum/blob/1c079bd7af4141bf7f86675b32f4168a6fa7939e/runtime/common/src/asset_registry.rs#L81).

runtime/common/src/lib.rs
```diff
@@ -184,3 +191,50 @@ impl<AssetRegistry: Inspect<AssetId = CurrencyId,Balance = Balance, CustomMetada
 			.ok_or(location)
 	}
 }
+
+
+pub struct OraclePriceGetter<Runtime>(PhantomData<Runtime>);
+impl < Runtime: treasury_buyout_extension::Config + dia_oracle::Config + orml_asset_registry::Config<AssetId = CurrencyId, CustomMetadata = CustomMetadata> > PriceGetter<CurrencyId> for OraclePriceGetter<Runtime> 
+{
```
ebma, line 198: The formatting here looks a bit off. Does the code compile?

bogdanS98: Yes, it compiles. I added another commit with formatting 👍🏼

ebma (approved):
LGTM!

PR #474, opened 2024-05-17
469 use asset registry metadata for implementation of diaoraclekeyconvertor

ebma:
Amendment to #469. Fixes a little oversight that was introduced in #470.


bogdanS98 (approved):
Looks good to me, nice catch! 👍🏼

PR #485, opened 2024-07-04
Add workflow to automatically create new node release

### LINKED ISSUE #456: Implement `--version` command
Ensure that the node clients return the version number, see this message from a third party collator:

> is it possible to have versioning on collators binaries? so when we use ./amplitude-collator --version we get the correct node version and also it will be displayed on telemetry correctly, so you can check for every provider which version is running

b-yap:
Partially closes #456. 

This workflow will do the following:
1. Perform a version up of the node
2. Create a new branch of the Cargo.toml change
3. Create a pull request with labels
   * if all packages are updated, the labels will be: `release-node`, `release-amplitude`, and `release-pendulum`
   * if only the node package is updated, the label will only be `release-node` 

### As we've never had a client release before, I've set this up to be triggered via the workflow. Like this:

Screenshot 2024-10-09 at 6 10 49 PM

### The PR will look like this:

Screenshot 2024-10-09 at 6 09 11 PM


gianfra-t (approved):
LGTM ✅. We need to be aware of this workflow to trigger it manually when we change the node then?

b-yap:
@gianfra-t Yes. 
Ideally after this workflow, it would be great to create a tag and then the release notes (which will be in gitlab, as is our practice). But I do not think it's the priority right now, hence the title **_[DO NOT MERGE]_**

ebma:
This is an interesting change for sure. I think the word 'workflow' in [this](https://github.com/pendulum-chain/pendulum/issues/456#issuecomment-2077840740) comment was not necessarily referring to a Github workflow but why not 😅 

The problem I see with this is that it doesn't mention the runtime in the title. For the other workflow, we [implemented](https://github.com/pendulum-chain/pendulum/blob/29e540a786ebc00b2ba8c6c876479e5a04a6ab34/.github/workflows/release.yml#L24-L37) it such that the title needs to contain 'release:' + either 'amplitude' or 'pendulum'. This is missing in this node version bump title. Now the question is how do we want to decide whether it should be for amplitude or pendulum? If we look at our usual flow, our release cycle is Foucoco -> Amplitude -> Pendulum. So this node version bump actually needs to happen when we upgrade the runtime on Foucoco. Our automated runtime upgrade workflow does not cover this, though, as we didn't include Foucoco in that. I think we could do one of the following:

1. Add Foucoco to the [release](https://github.com/pendulum-chain/pendulum/blob/main/.github/workflows/release.yml) workflow and change the version-workflow to always define `release: ... Foucoco` in the title. Then the node version is always bumped in time.
2. Remove `release:` from the title of the PRs automatically created by this workflow and then we need to take care of the rest manually as per usual. 
 
Since so far, we don't really have a well-defined process for creating and deploying the node upgrade in the first place, maybe option 1) is preferable. WDYT @pendulum-chain/devs?

gianfra-t:
So if I understand idea 1 correctly, once we perform a change to the node and trigger this version-workflow, it will also trigger a Foucoco release. But couldn't the node and runtime upgrades be independent of each other? Because we may update the node but not make any change to the runtime itself I assume.

ebma:
> But couldn't the node and runtime upgrades be independent of each other? Because we may update the node but not make any change to the runtime itself I assume.

While this is true, I think this is rarely the case and most of the time, we'll only want to ship a node upgrade in combination with a runtime upgrade. So maybe it's better not to handle that edge case for the sake of simplicity.

b-yap:
@ebma  @gianfra-t  
From what I remember, Node upgrade is independent from the runtime release, and should be performed on **all** runtimes. https://satoshipay.slack.com/archives/C047E2XQ160/p1710232364946679

We have never defined "foucoco release"; what's the criteria?
Also, adding a specific "foucoco release" in a PR title will seem as if the node upgrade is _only_ for foucoco. 

The upcoming [polkadot-v1.0.0](https://github.com/pendulum-chain/spacewalk/pull/536) will change the version of the spacewalk dependency in `pendulum-node`, and upgrade is for all runtimes. I think a PR that states "node release" is appropriate to trigger this, right? Not necessarily Foucoco.

gianfra-t:
> I think a PR that states "node release" is appropriate to trigger this, right? Not necessarily Foucoco.

So it then would be independent of the runtime? I would prefer this. I understand the last @ebma comment as follows: we will (almost) always make a modification of the node code while also doing so on the runtime, which will generally be Foucoco.

But, once the node code has been bumped (and changed), should we also bump it's version for every other consequent runtime upgrade? I understood differently.

ebma:
I checked other parachain teams. 

Moonbeam has extra releases for just the collator nodes however. [This](https://github.com/moonbeam-foundation/moonbeam/releases/tag/v0.38.0) is an example of a node release and [this](https://github.com/moonbeam-foundation/moonbeam/releases/tag/runtime-3000) of the runtime.

From what I can tell, centrifuge only have the runtimes in their [releases](https://github.com/centrifuge/centrifuge-chain/releases). 

Astar add their collators to the assets of each [release](https://github.com/AstarNetwork/Astar/releases) and so does [interlay](https://github.com/interlay/interbtc/releases). 
We cannot apply this to us though as they group their runtime releases and do one release for all at the same time, while we have separate releases for Amplitude and Pendulum.

b-yap:
@ebma We can definitely follow moonbeam. I'll check their workflow and update this PR next week.

ebma:
Cool 👍 

@pendulum-chain/devs any objections if in addition to our current workflow, we share upgrades to the collator nodes in separate releases on Github?

TorstenStueber:
So if I understand this right with this PR we would have the following process:
* a manual git workflow to create a PR that simply bumps the Cargo.toml version of the node
* this is then independent of GitHub releases for runtime versions (for Amplitude and Pendulum separately) we create mostly manually

I think it makes sense to keep this separate conceptually as the node release applies to all runtimes at the same time – even if we will practically mostly only update the node version together with at least one runtime upgrade.

Shall we also create a GitHub release whenever we upgrade the node version or at least create a tag for the merge commit of the PR?

ebma:
Yes, I think we should put the node upgrades into separate Github releases similar to what Moonbeam does.

b-yap:
@ebma @TorstenStueber 
We'll need @zoveress to setup the release on gitlab. 

I've setup the get_commit.sh, similar to the `runtime-upgrade-automation`. 
We just have to decide the assets to attach; 
* pendulum-node binary
* source code of only the node? Or the entire pendulum?

.github/workflows/node-version-up.yml
```diff
@@ -0,0 +1,63 @@
+name: Node Version Up
+
+on:
+  workflow_dispatch:
+    inputs:
+      versioning:
+        description: "node version up: major, minor, or patch"
```
ebma, line 7: Does this allow setting a custom version? If I want to set the version to `1.1.0-beta`, would I be able to pass this here and it's applied like that in the Cargo.toml?

.github/workflows/node-version-up.yml
```diff
@@ -0,0 +1,63 @@
...
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+      - name: install cargo-edit
+        run: |
+          cargo install cargo-edit
+      - name: Set New Version ${{ github.event.inputs.versioning }}
+        continue-on-error: false
+        run: |
+          cargo set-version --bump ${{ github.event.inputs.versioning }} --package pendulum-node
```
ebma, line 30: I would prefer to change the version of all crates but this is to be seen after https://github.com/pendulum-chain/pendulum/pull/493.

.github/workflows/node-version-up.yml
b-yap: So, cargo edit allows 6 options: [major, minor, patch, release, rc, alpha, beta](https://github.com/killercup/cargo-edit/blob/f7ad38c74c03e9cb6dab3eeb82601a88ef2f2c06/src/bin/set-version/version.rs#L60-L67). 

Example of choosing alpha: `version = "0.1.3-alpha.1"`

.github/workflows/node-version-up.yml
b-yap: I've added 3 choices: 
node, all packages, or all packages but the node.

.github/workflows/version-up.yml
```diff
@@ -0,0 +1,186 @@
...
+
+      - name: Create Pull Request
+        uses: thomaseizinger/create-pull-request@master
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          title: "release: ${{ github.event.inputs.versioning }} version up ${{ steps.set-pkg.outputs.name }}"
+          body: "${{ steps.read_file.outputs.content }}"
+          head: ${{ steps.new-branch.outputs.name }}
+          base: main
+          reviewers: "ebma, TorstenStueber, bogdanS98, gianfra-t, b-yap"
```
ebma, line 186: Let's remove the hard-coded reviewers. We can just make it use '@pendulum-chain/devs' manually once the PR is created.

.github/workflows/node-version-up.yml
ebma: That's nice 👍 

I think there should also be the option to set an [absolute version](https://github.com/killercup/cargo-edit/blob/f7ad38c74c03e9cb6dab3eeb82601a88ef2f2c06/src/bin/set-version/version.rs#L30) instead of bumping it. I guess we'd need to use that more often, e.g. looking at the next bump from v1.1.0 to v1.6.0, we can't use any of the relative bumps as we want to skip a couple versions at once. I think in that case we can just do it with `cargo set-version 1.6.0` ie without the `--bump` parameter.

.github/workflows/node-version-up.yml
ebma: I think we don't need to be able to choose between the three. It's nice-to-have but maybe better to remove since you had to duplicate large parts of it.

ebma:
> I've setup the get_commit.sh, similar to the runtime-upgrade-automation.
We just have to decide the assets to attach;
pendulum-node binary
source code of only the node? Or the entire pendulum?

I think it's fine if we just attach the compiled node binary to the release. The .zip file of the source code is always attached to a Github release anyway, so nothing to add here. Let's coordinate with @zoveress next week.

.github/workflows/node-version-up.yml
b-yap: So how do you want to upgrade? Just between node or ALL?

ebma: Let's update all crates by default.

.github/workflows/node-version-up.yml
b-yap: Can't do it in `set-version`. I've copied [the script from spacewalk](https://github.com/pendulum-chain/spacewalk/blob/main/.maintain/create_minor_release.sh) that requires a _version_ argument.
example: `.github/scripts/specific_version_up 1.6.0`

.github/workflows/node-version-up.yml
b-yap: default set to true.

.github/workflows/version-up.yml
b-yap: updated it.

.github/workflows/release-node.yml
```diff
@@ -0,0 +1,31 @@
...
+      - closed
+    branches:
+      - 'main'
+
+jobs:
+  release_check:
+#    This job will only run if:
+#     * the pull request is closed and merged to main branch;
+#     * the pull request has the "release:" and "node" in its title
+    if: ${{ github.event.pull_request.merged == true && contains(github.event.pull_request.title, 'release:') && contains(github.event.pull_request.title, 'node')  }}
```
ebma, line 16: We would need to make sure that the PRs now include both `release:` and `node`. The PRs you automatically create in the version-up workflow don't do this. How about we change the condition here to check for `release node: ....` instead of checking for both words separately and then we make the version-up workflow create the PR title `release node: ` prefixed?

b-yap: edited

.github/workflows/version-up.yml
```diff
@@ -0,0 +1,174 @@
...
+          -F commitMessage="$(cat changes.txt)" \
+          -F files[][path]="node/Cargo.toml" -F files[][contents]=$(base64 -w0 node/Cargo.toml) \
+          -F 'query=@.github/api/createCommitOnBranch.gql'
+
+      - name: Create Pull Request for Specific Version
+        if: github.event.inputs.specific_version != ''
+        uses: thomaseizinger/create-pull-request@master
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          title: "release: Force version up ${{ steps.set-pkg.outputs.name }}"
```
ebma, line 159:
```suggestion
          title: "release node: Force version up ${{ steps.set-pkg.outputs.name }}"
```

.github/workflows/version-up.yml
```diff
@@ -0,0 +1,174 @@
...
+          head: ${{ steps.new-branch.outputs.name }}
+          base: main
+          reviewers: "pendulum-chain/devs"
+
+      - name: Create Pull Request
+        if: github.event.inputs.specific_version == ''
+        uses: thomaseizinger/create-pull-request@master
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          title: "release: ${{ github.event.inputs.versioning }} version up ${{ steps.set-pkg.outputs.name }}"
```
ebma, line 170:
```suggestion
          title: "release node: ${{ github.event.inputs.versioning }} version up ${{ steps.set-pkg.outputs.name }}"
```

ebma (commented):
We should apply that to the automatic title as well then.

.github/workflows/version-up.yml
b-yap: Okay so instead of the PR title, I've used labels instead. I think it's less prone to mistakes.
I have updated the description of the PR

.github/workflows/version-up.yml
```diff
@@ -0,0 +1,181 @@
...
+
+      - name: Prepare Pull Request title and labels
+        id: pr-title-labels
+        # if all packages are updated, then add all labels
+        run: |
+          if ${{ github.event.inputs.packages-all }} == 'true'; then
+            echo "label=release-node,release-amplitude,release-pendulum" >> "$GITHUB_OUTPUT"
+          else
+            echo "label=release-node" >> "$GITHUB_OUTPUT"
+          fi
```
ebma, line 164: Since we will not use the `release-amplitude` and `release-pendulum` label, let's just stick to the `release-node` in all cases. At least I don't see a benefit of these other two labels. Unless we use those labels for the runtime upgrades too, instead of the `release:`. But that's for another ticket on another day.

b-yap: Alright I'll remove them both.

ebma (approved):
Looks good to me 👍  let's coordinate with Zoltan regarding the node creation pipeline before merging this though.

ebma (approved):

ebma:
The actions added to this PR were tested [here](https://github.com/b-yap/pendulum/actions).

PR #487, opened 2024-07-19
Implement pallet for mock-skip blocks on runtimes

### LINKED ISSUE #486: Implement and test potential solution for mock-skip blocks on runtimes
Continuation from this [ticket](https://github.com/pendulum-chain/tasks/issues/297).
This ticket comes directly as an action point from #190.

### Context and recap
We are currently using an isolated foucoco-based [runtime](https://github.com/pendulum-chain/foucoco-standalone) for wasm-deploy testings. This variant of the runtime allows us to skip blocks for testing purposes by calling a modified pallet contracts which in turns modifies the reported block number that the contracts see.

We want to remove this extra repository and have this functionality on the main Foucoco [repository](https://github.com/pendulum-chain/pendulum/tree/main/runtime) itself.

### Solution to explore 

We would like to create a new pallet that would be conditionally compiled with a `test` feature flag (or equivalent). This pallet will have at least one extrinsic, which will not take fee nor check signature, that will set the desired block we would like the chain to report during contracts execution. 

For this, we can leverage the `on_initialize` hook. When called, we will store the current block number and then call directly the `set_block_number` [method](https://github.com/paritytech/substrate/blob/master/frame/system/src/lib.rs#L1524C9-L1524C25) of the frame system pallet and set it to the desired value. 

Then, when `on_finalize` is called, this pallet should revert the changes done on the frame system pallet. The idea is that only during execution of extrinsic the block reported is different to what it should actually be, and does not interfere with production of blocks. See the [lifecycle](https://paritytech.github.io/polkadot-sdk/master/frame_support/traits/trait.Hooks.html#summary) of hooks and extrinsics for context.

Since this will likely introduce an issue with the validity of any transaction made whenever the block has been changed (see this comment for more [details](https://github.com/pendulum-chain/tasks/issues/190#issuecomment-2033196981)), we need to modify the instances were wasm-deploy submits extrinsics to the chain such that the transaction is specified as `Immortal`. The calls are formed [here](https://github.com/pendulum-chain/api-solang/blob/fdf4a2870f3bf3639dd85b53c62ae2939df65214/src/deployContract.ts#L64) and [here](https://github.com/pendulum-chain/api-solang/blob/fdf4a2870f3bf3639dd85b53c62ae2939df65214/src/index.ts#L215-L221) either for deployment or call to an existing contract. Since `api-solang` already allows for specifying a `modifyExtrinsic` [function](https://github.com/pendulum-chain/api-solang/blob/fdf4a2870f3bf3639dd85b53c62ae2939df65214/src/index.ts#L174), this would be a good place modify the transaction from `Mortal` to `Immortal`.

### Testing
Ultimately, we are currently using the functionality to skip blocks [here](https://github.com/pendulum-chain/wasm-deploy/blob/572e60dd08a9de5172a2d5e750a74fd17963e5d1/src/commands/test.ts#L346). Once the new solution is in place, we should be able to modify the `skipBlock` function so it calls the new pallet and test if the solution works.

bogdanS98:
This PR introduces a new pallet used to mock block skipping when testing wasm-deploy.
The idea behind it is described in [this](https://github.com/pendulum-chain/pendulum/issues/486) ticket in detail. 
I'll add a broader description before marking the PR as ready for review.

Closes #486.

### DIFF
--- a/pallets/pallet-mock-skip-blocks/src/lib.rs
+++ b/pallets/pallet-mock-skip-blocks/src/lib.rs
@@ -1,0 +1,110 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+#[cfg(feature = "instant-seal")]
+pub use pallet::*;
+
+#[cfg(not(feature = "instant-seal"))]
+pub use dummy as pallet;
+
+#[cfg(not(feature = "instant-seal"))]
+#[frame_support::pallet]
+pub mod dummy {
+	use frame_support::pallet_prelude::*;
+
+    #[pallet::pallet]
+    pub struct Pallet<T>(_);
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {        
+        /// Dummy event
+       DummyEvent,
+	}
+
+    #[pallet::config]
+    pub trait Config: frame_system::Config {
+		/// Overarching event type
+        type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+	}
+
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {}
+}
+
+#[cfg(test)]
+mod mock;
+
+#[cfg(test)]
+mod tests;
+
+#[cfg(feature = "instant-seal")]
+#[frame_support::pallet]
+pub mod pallet {
+    use frame_support::{dispatch::DispatchResult, pallet_prelude::*};
+    use frame_system::pallet_prelude::*;
+
+    #[pallet::pallet]
+    pub struct Pallet<T>(PhantomData<T>);
+
+    #[pallet::error]
+    pub enum Error<T> {
+        InvalidBlockNumber,
+    }
+
+    #[pallet::event]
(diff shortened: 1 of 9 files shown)


pallets/pallet-mock-skip-blocks/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {}
+}
+
+#[cfg(test)]
+mod mock;
+
+#[cfg(test)]
+mod tests;
```
ebma, line 38: If the whole content of mock and test are only supposed to be used when `instant-seal` is available, wouldn't it make more sense to instead move the feature flag macro above the `mod` imports here? Then we can remove all the annotations from the `mock.rs` and `test.rs` files itself.

bogdanS98: True 👍🏼  I forgot about those macro flags when testing some things earlier

ebma: Maybe we can even move them into the `pub mod pallet {}` definition? Not sure if that works but then no extra macros were needed.

PR #488, opened 2024-07-24
420 change github workflow to run test with feature flags

### LINKED ISSUE #420: Change Github workflow to run tests with feature flags
### Context

At the moment, our Github workflow is running `cargo test` without any feature flags. This allows for changes that break benchmarks to pass the pipeline and therefore be merged into the main branch. This will cause building the main branch with `runtime-benchmarks` feature to fail. The same thing happens for pallets/runtimes (i.e. development runtime) that do not properly declare the `try-runtime` feature.

### Requirement

Pipeline should be changed so that it runs `cargo test --features runtime-benchmarks try-runtime`

### Notes

Before changing the workflow to run tests with the `try-runtime` feature, development runtime must be removed since it's not used anymore and it's causing errors when built with `try-runtime`. There's already an open ticket for deleting this runtime [here](https://github.com/pendulum-chain/pendulum/issues/80).

b-yap:
closes #420 
Add another step to run a test but with features `runtime-benchmarks` and `try-runtime`.

This also removes `nightly`


.github/workflows/test-code.yml
```diff
@@ -45,6 +45,13 @@ jobs:
           toolchain: nightly-2024-05-30
           command: test
 
+      - name: Test For runtime benchmarks
+        uses: actions-rs/cargo@v1
+        with:
+          toolchain: nightly-2024-05-30
+          command: test
+          args: --all --features runtime-benchmarks try-runtime
```
ebma, line 53: Don't we need to wrap it in quotation marks?
```suggestion
          args: --all --features "runtime-benchmarks try-runtime"
```

Also, can't we remove the other test in ll. 42-46 then?

ebma:
Just so I understand properly, you are saying that the change requested in #420 is useless because all tests are filtered out? 

Let's maybe change the workflow [here](https://github.com/pendulum-chain/pendulum/pull/488/files#diff-9d8a210f3bc0ae57dbd26c81f07a2811ad8b593f2b125e5b37e37522f90ff644R3-R6) to run the checks on any PR, not just on PRs that merge to `main`.

Cargo.toml
```diff
@@ -33,5 +33,6 @@ orml-tokens = { git = "https://github.com/open-web3-stack//open-runtime-module-l
 ### orml packages { version = "0.4.1-dev" }
 ### zenlink packages  { version = "*" }
 
-[patch."https://github.com/w3f/ring-vrf"]
-bandersnatch_vrfs = { git = "https://github.com/w3f//ring-vrf", version = "0.0.4", rev = "3ddc20" }
\ No newline at end of file
+# probable patch for nightly
```
TorstenStueber, line 36: Is this not required anymore? In that case: can we remove it (instead of commenting it out)?

TorstenStueber (commented):
Similar to @ebma's question, to me the intention of https://github.com/pendulum-chain/pendulum/issues/420 is not entirely clear: is the task to 1) just filter out all test for `--features runtime-benchmarks` or is it to 2) make these tests work also for `--features runtime-benchmarks`.

The issue says:
> This allows for changes that break benchmarks to pass the pipeline and therefore be merged into the main branch

This almost sounds to me like it means option 2.

Cargo.toml
bogdanS98: My intention when I wrote the ticket was that whatever cargo command the pipeline runs (which was just plain `cargo test` when I wrote the ticket) it would be ran using the `runtime-benchmarks` feature so that the pipeline would catch any benchmarks related errors. So this is more like option 2 in my opinion.

ebma (approved):
It's great to remove the need for stable vs nightly configurations and it's also nice that you extracted the clippy checks from the tests so they can already run in parallel. 

Ready for merge IMO 👍

PR #489, opened 2024-08-01
WIP Polkadot v1.1.0

### LINKED ISSUE #481: Update to Polkadot version 1.1.0 - Pendulum
Related to [tasks/#306](https://github.com/pendulum-chain/tasks/issues/306)

We need to perform the upgrade of Pendulum to [polkadot-sdk release v1.1.0](https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.1.0).

This involves updating all corresponding dependencies in this repository:
- Self-maintained pallets
- Node client
- Chain extension code
- Runtimes (Including storage migrations)

**Before performing the runtime upgrade** with this changes it is essential to have [checked](https://github.com/pendulum-chain/pendulum/issues/482) which are the necessary migrations and added the migration code, if needed.

gianfra-t:
Please ensure that the collator nodes are redeployed after this PR is merged.  

Closes #481 .


prayagd:
> redeployment of the collator nodes to take effect.

Do we have a ticket for this, so i can link here as blocked?

gianfra-t:
@prayagd they have to be redeployed but after the changes on this PR are finished. Sorry it pinged you since it is still a draft.

rustc-ice-2024-08-19T10_33_10-40406.txt
```diff
@@ -0,0 +1,599 @@
+thread 'rustc' panicked at compiler/rustc_middle/src/ty/normalize_erasing_regions.rs:187:90:
```
b-yap, line 1: Are you including this file in PR?

runtime/amplitude/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 // `construct_runtime!` does a lot of recursion and requires us to increase the limit to 256.
```
b-yap, line 2: limit to 512.

b-yap, line 2: limit to 512; or just remove "to 256"

rustc-ice-2024-08-19T10_33_10-40406.txt
gianfra-t: That one sneaked in probably. Need to remove it.

runtime/amplitude/src/lib.rs
```diff
@@ -609,10 +565,15 @@ impl pallet_session::Config for Runtime {
 	type WeightInfo = pallet_session::weights::SubstrateWeight<Runtime>;
 }
 
+parameter_types! {
+	// as per documentation, typical value for this is false "unless this pallet is being augmented by another pallet"
```
b-yap, line 569: Can you add the documentation link, here?

runtime/integration-tests/Cargo.toml
```diff
@@ -10,54 +10,71 @@ codec = { package = "parity-scale-codec", version = "3.0.0" }
 scale-info = { version = "2.1.2", features = ["derive"] }
 serde = { version = "1.0.144", features = ["derive"] }
 paste = "1.0.14"
+log = { version = "0.4.17" }
```
b-yap, line 13: OR, can we just remove the `trace!` code in line 332 of sibling.rs? 

I was thinking, that we really don't need any more logs in this testing package

runtime/amplitude/src/lib.rs
gianfra-t: Of course!

runtime/integration-tests/Cargo.toml
gianfra-t: You're right, we don't even have the log enabled. We took this from the runtime's xcm configs as-is in the past probably.

runtime/integration-tests/src/amplitude_tests.rs
```diff
@@ -12,95 +11,145 @@ use crate::{
...
+	BASE_FEE_WHEN_TRANSFER_NON_NATIVE_ASSET / 10;
 
-decl_test_relay_chain! {
-	pub struct KusamaRelay {
-		Runtime = kusama_runtime::Runtime,
-		XcmConfig = kusama_runtime::xcm_config::XcmConfig,
-		new_ext = kusama_relay_ext(),
-	}
+decl_test_relay_chains! {
+	#[api_version(5)]
```
ebma, line 33: How do we know it's `5`?

runtime/integration-tests/src/amplitude_tests.rs
```diff
@@ -12,95 +11,145 @@ use crate::{
...
+		Kusama,
 		amplitude_runtime,
-		AmplitudeParachain
+		AmplitudeParachain,
+		AMPLITUDE_ID,
+		KSM_FEE_WHEN_TRANSFER_TO_PARACHAIN
 	);
 }
-
+//
```
ebma, line 147: Nit: remove

runtime/integration-tests/src/genesis.rs
```diff
@@ -0,0 +1,144 @@
...
+
+
+#[macro_export]
+macro_rules! genesis_gen {
+	($runtime:ident, $para_account_id: ident, $asset_metadata: ident) => {{
+		use $runtime::BuildStorage;
+        use crate::mock::units;
+        use integration_tests_common::constants::{collators, accounts};
+        use spacewalk_primitives::CurrencyId;
+        pub const SAFE_XCM_VERSION: u32 = 3;
```
ebma, line 23: Do we need to define this twice? It's already defined in l. 10.

runtime/integration-tests/src/sibling.rs
```diff
@@ -311,10 +328,7 @@ impl ShouldExecute for DenyReserveTransferToRelayChain {
 		if matches!(origin, MultiLocation { parents: 1, interior: Here }) &&
 			instructions.iter().any(|inst| matches!(inst, ReserveAssetDeposited { .. }))
 		{
-			log::trace!(
-				target: "xcm::barriers",
-				"Unexpected ReserveAssetDeposited from the relay chain",
-			);
+			println!{"Unexpected ReserveAssetDeposited from the relay chain"};
```
ebma, line 331: Why do we change it away from log::trace?

runtime/integration-tests/src/test_macros.rs
```diff
@@ -8,39 +8,59 @@ macro_rules! transfer_20_relay_token_from_relay_chain_to_parachain {
...
 
 		// a "Deposited" event occurred is proof that the transfer was successful
 		$parachain::execute_with(|| {
 			use $para_runtime::{RuntimeEvent, System};
+
+			for event in System::events() {
+				println!{"Parachain events when receiving of relay token"}
+				println!{"{:?}", event}
+			}
+
```
ebma, line 63: Should we remove these logs again?

runtime/integration-tests/src/test_macros.rs
```diff
@@ -117,7 +149,11 @@ macro_rules! transfer_10_relay_token_from_parachain_to_relay_chain {(
 			use $relay_runtime::{RuntimeEvent, System, Balances};
 
 			let events = System::events();
-			assert_eq!(events.len(), 3);
+
+			for event in System::events() {
+				println!{"Receiveing relay token on relay chain"}
+				println!{"{:?}", event}
+			}
```
ebma, line 157: Same here

runtime/integration-tests/Cargo.toml
```diff
@@ -12,52 +12,68 @@ serde = { version = "1.0.144", features = ["derive"] }
...
+orml-xcm-support = { git = "https://github.com/open-web3-stack/open-runtime-module-library.git", branch = "polkadot-v1.1.0" }
+orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library.git", branch = "polkadot-v1.1.0" }
+orml-tokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library.git", branch = "polkadot-v1.1.0" }
+orml-xtokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v1.1.0" }
+
+pallet-treasury = { git = "https://github.com/paritytech/polkadot-sdk", branch = "release-polkadot-v1.1.0" }
 
 # Local
-runtime-common = { path = "../common", default-features = false }
+runtime-common = { path = "../common" , default-features = false }
```
ebma, line 76:
```suggestion
runtime-common = { path = "../common", default-features = false }
```

ebma (commented):
Thanks for all the great changes @gianfra-t 👌 I checked most parts and it looks good to me. I left a lot of comments but most of them are questions or minor suggested changes.

> The solution to this issue was to implement a patch with all packages specified to normalize for this branch (attempted in this [commit](https://github.com/pendulum-chain/pendulum/pull/489/commits/ad33520dcfa01221163d3857190e9f35b6c10ea0#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R43))

I'm not sure I understand this part. I can see in the commit you linked to that you patched all the crates. But these patches are not contained in the most recent version anymore. Does this mean you want to rely on the current state of the `Cargo.lock` file? Seems a little brittle to me. Or am I missing something here?

runtime/integration-tests/src/amplitude_tests.rs
b-yap: Yeah, I think the [latest api version is 11](https://github.com/paritytech/polkadot-sdk/blob/f6eeca91b6cf0810ca3d8e7ea23988d9510851ba/polkadot/primitives/src/runtime_api.rs#L296). But for `release-polkadot-v1.1.0`, I think [version 5 is enough](https://github.com/paritytech/polkadot-sdk/blob/release-polkadot-v1.1.0/polkadot/primitives/src/runtime_api.rs#L129).

runtime/common/src/benchmarking/orml_asset_registry.rs
```diff
@@ -76,14 +78,14 @@ pub mod benchmarks {
 			RawOrigin::Root,
 			CurrencyId::Native,
 			Some(123),
-			Some(vec![b'b', 128]),
-			Some(vec![b'b', 128]),
+			Some(BoundedVec::truncate_from(vec![b'b', 128])),
```
b-yap, line 81: So, the reason for not using `longest_vec::<T>()` here, is the change from `b'a'` to `b'b'` ?

runtime/amplitude/src/lib.rs
```diff
@@ -148,62 +151,42 @@ pub type SignedPayload = generic::SignedPayload<RuntimeCall, SignedExtra>;
...
+use runtime_common::asset_registry::StringLimit;
 
 parameter_types! {
 	pub const InactiveAccounts: Vec<AccountId> = Vec::new();
 }
 
-// Temporary struct that defines the executions to be done upon upgrade,
-// Should be removed or at least checked on each upgrade to see if it is relevant,
-// given that these are "one-time" executions for particular upgrades
+// TODO remove contract migrations after update
```
b-yap, line 160: After update of? Version 1.1.0?

pallets/parachain-staking/src/set.rs
```diff
@@ -16,10 +16,10 @@
 
 // If you feel like getting in touch with us, you can do so at info@botlabs.org
 
-use frame_support::{traits::Get, BoundedVec, DefaultNoBound, RuntimeDebug};
+use frame_support::{traits::Get, BoundedVec, DefaultNoBound};
 use parity_scale_codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
-use sp_runtime::{traits::Zero, SaturatedConversion};
+use sp_runtime::{traits::Zero, SaturatedConversion, RuntimeDebug};
```
b-yap, line 22: Ooh, `RuntimeDebug` can be accessed via `sp_runtime`.

So we can actually:
1. Remove `sp-debug-derive` dependency in runtime/integration-tests/Cargo.toml;
2. Then replace `use sp_debug_derive::RuntimeDebug;` in runtime/integration-tests/src/sibling.rs with 
`use sp_runtime::RuntimeDebug`

node/src/chain_spec.rs
```diff
@@ -1,4 +1,7 @@
+#![allow(clippy::default_constructed_unit_structs)]
+
 use cumulus_primitives_core::ParaId;
+use frame_benchmarking::__private::sp_std;
```
b-yap, line 4: Use `sp_std` from **`sp_core`** instead.
Let's not call from private, even though it's public.

runtime/integration-tests/src/sibling.rs
b-yap: This is to remove the log dependency, as it was only added because of this.
I believe the less dependencies we have, the better for us to manage version ups.
I think a removal of even 1 dependency is a win :raised_hands:

Cargo.toml
```diff
@@ -25,21 +25,13 @@ members = [
...
+orml-traits = { git = "https://github.com/open-web3-stack//open-runtime-module-library", branch = "polkadot-v1.1.0" }
+orml-tokens = { git = "https://github.com/open-web3-stack//open-runtime-module-library", branch = "polkadot-v1.1.0" }
+##
+### need this because of bifrost farming dependency in runtime
+### bifrost uses :
+### orml packages { version = "0.4.1-dev" }
+### zenlink packages  { version = "*" }
 
+[patch."https://github.com/w3f/ring-vrf"]
+bandersnatch_vrfs = { git = "https://github.com/w3f//ring-vrf?rev=3ddc20", version = "0.0.4", rev = "3ddc20" }
```
b-yap, line 37: So... I could run: 
```
cargo b -p amplitude-runtime --features runtime-benchmarks
``` 
in `main` branch with no problems.

But now:
```
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
note: for more details see https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions
    Updating git repository `https://github.com/w3f//ring-vrf?rev=3ddc20`
fatal: https://github.com/w3f//ring-vrf?rev=3ddc20/info/refs not valid: is this a git repository?
error: failed to load source for dependency `bandersnatch_vrfs`

Caused by:
  Unable to update https://github.com/w3f//ring-vrf?rev=3ddc20?rev=3ddc20

Caused by:
  failed to fetch into: /Users/b.carlayap/.cargo/git/db/ring-vrf-5f0e10e9eb5881f6
```

If you encountered this, do you know the solution?

ebma:
```
cargo b -p amplitude-runtime --features runtime-benchmarks
```
works for me on `polkadot-v1.1.0`. Did you `cargo clean` before building on the other branch?

gianfra-t: Yes, please run `cargo clean` or `cargo update -p bandersnatch_vrfs`, since probably the patch needs to be applied on your local setup @b-yap.

gianfra-t:
@ebma about this question:

> But these patches are not contained in the most recent version anymore. Does this mean you want to rely on the current state of the Cargo.lock file? Seems a little brittle to me. Or am I missing something here?

We are not relying on the lock file, we required the patches if we were to use the assethub runtimes from the `polkadot-fellows` [repo](https://github.com/polkadot-fellows/runtimes). But if we use the assethub runtimes from the `polkadot-sdk` [directly](https://github.com/pendulum-chain/pendulum/pull/489/files#diff-64f76b6252402cf0f4b321854794de1bb8dabb4712de81965fa98568648a0429R59), then there is no compilation issue. The difference as I put in the description is that these are [fixed](https://github.com/paritytech/polkadot-sdk/blob/release-polkadot-v1.1.0/cumulus/parachains/runtimes/assets/asset-hub-kusama/Cargo.toml#L3) at `0.9.42`, even though the `sdk` version is `1.1.0`. I believe they just made them compatible but did not upgrade the logic itself. So we have the choice to use either:
- Patch + runtimes from the `polkadot-fellows` repo.
- Runtimes from the `polkadot-sdk`

When I discovered the second option was also working, I chose that one since maintaining the patch is very brittle. Especially since we cannot do `cargo update` without breaking our lock.

About the weights, yes I didn't think about regenerating the weights again. Do you think it can be a bit too much for a single PR?

node/src/chain_spec.rs
```diff
@@ -488,15 +491,17 @@ fn amplitude_genesis(
...
 
-	amplitude_runtime::GenesisConfig {
+	amplitude_runtime::RuntimeGenesisConfig {
 		asset_registry: Default::default(),
 		system: amplitude_runtime::SystemConfig {
 			code: amplitude_runtime::WASM_BINARY
 				.expect("WASM binary was not build, please build it!")
 				.to_vec(),
+			#[allow(clippy::wrong_self_convention)]
+			_config: sp_std::marker::PhantomData::default()
```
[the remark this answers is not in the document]
gianfra-t: It was added (started being required) when we [removed](https://github.com/pendulum-chain/pendulum/pull/489/files) the specific import of the struct from the pallets, I think. I believe it is a small price to pay in terms of extra code to keep the `construct_runtime` cleaner. WDYT?

node/src/chain_spec.rs
```diff
@@ -564,6 +570,7 @@ fn amplitude_genesis(
 			} else {
 				amplitude_runtime::StatusCode::Error
 			},
+			_phantom: sp_std::marker::PhantomData::default()
```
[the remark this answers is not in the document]
gianfra-t: Again related to the previous [comments](https://github.com/pendulum-chain/pendulum/pull/489#discussion_r1741097619) but I don't know the exact reason.

node/src/command.rs
```diff
@@ -331,14 +314,9 @@ pub fn run() -> Result<()> {
...
-			let runner = cli.create_runner(cmd)?;
-			runner.sync_run(|_config| {
-				let spec = cli.load_spec(&cmd.shared_params.chain.clone().unwrap_or_default())?;
-				let state_version = Cli::native_runtime_version(&spec).state_version();
-				cmd.run::<Block>(&*spec, state_version)
-			})
-		},
+		Some(Subcommand::ExportGenesisState(cmd)) => construct_async_run!(|components, cli, cmd, config| {
+                Ok(async move { cmd.run(&*config.chain_spec, &*components.client) })
+            }),
```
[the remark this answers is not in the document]
gianfra-t: Will run again `cargo fmt`, perhaps this change entered after I did.

node/src/service.rs
```diff
@@ -688,51 +697,40 @@ where
...
-			}
-		},
+	let collator_service = CollatorService::new(
+		client.clone(),
+		Arc::new(task_manager.spawn_handle()),
+		announce_block.clone(),
+		client.clone(),
+	);
+
+	let _relay_chain_interface_move = relay_chain_interface.clone();
```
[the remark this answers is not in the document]
gianfra-t: Totally 🤦🏻‍♂️.

node/src/service.rs
```diff
@@ -416,9 +434,9 @@ where
 	)
 	.await
 	.map_err(|e| sc_service::Error::Application(Box::new(e)))?;
-	let block_announce_validator = BlockAnnounceValidator::new(relay_chain_interface.clone(), id);
+	let block_announce_validator = RequireSecondedInBlockAnnounce::new(relay_chain_interface.clone(), id);
```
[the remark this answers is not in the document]
gianfra-t: It has been [renamed](https://github.com/paritytech/polkadot-sdk/blob/release-polkadot-v1.1.0/cumulus/client/network/src/lib.rs#L190), no functionality change.

runtime/amplitude/src/lib.rs
```diff
@@ -148,62 +151,42 @@ pub type SignedPayload = generic::SignedPayload<RuntimeCall, SignedExtra>;
...
+
+		if Bounties::on_chain_storage_version() == 0 {
+			log::info!("Upgrading bounties storage version to 4");
+			StorageVersion::new(4).put::<Bounties>();
 			writes += 1;
 		}
-
-		<Runtime as frame_system::Config>::DbWeight::get().reads_writes(7, writes)
+		// not really a heavy operation
+		<Runtime as frame_system::Config>::DbWeight::get().reads_writes(4, writes)
```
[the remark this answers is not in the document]
gianfra-t: Yes! I forgot to change it from the last upgrade.

runtime/amplitude/src/lib.rs
```diff
@@ -275,7 +256,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("amplitude"),
 	impl_name: create_runtime_str!("amplitude"),
 	authoring_version: 1,
-	spec_version: 17,
+	spec_version: 18,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 13,
```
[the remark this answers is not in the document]
gianfra-t: I thought about it, I deliberately left it unchanged but we have to discuss it, since changing it will brake integrations, but not changing it may also cause confusion if the calls have changed substantially.

runtime/amplitude/src/lib.rs
```diff
@@ -1005,6 +996,9 @@ parameter_types! {
 	pub const DepositPerByte: Balance = deposit(0, 1);
 	pub const DefaultDepositLimit: Balance = deposit(1024, 1024 * 1024);
 	pub Schedule: pallet_contracts::Schedule<Runtime> = Default::default();
+	// TODO is this value okay?
+	pub const CodeHashLockupDepositPercent: Perbill = Perbill::from_percent(10);
```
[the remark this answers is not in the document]
gianfra-t: I saw [it](https://github.com/AstarNetwork/Astar/blob/f5c99e9b2ee41ccae068516109c0e0c97f2eceab/runtime/local/src/lib.rs#L711) in Astar. In any case, it is something we need to choose for our particular business logic.

runtime/amplitude/src/lib.rs
```diff
@@ -568,6 +550,8 @@ impl cumulus_pallet_parachain_system::Config for Runtime {
 	type XcmpMessageHandler = XcmpQueue;
 	type ReservedXcmpWeight = ReservedXcmpWeight;
 	type CheckAssociatedRelayNumber = RelayNumberStrictlyIncreases;
+	#[cfg(feature = "std")]
+	type ConsensusHook = cumulus_pallet_parachain_system::consensus_hook::ExpectParentIncluded;
```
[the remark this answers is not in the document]
gianfra-t: This is actually an odd one. Now, this type is only supposed to be required when [this flag](https://github.com/paritytech/polkadot-sdk/blob/release-polkadot-v1.1.0/cumulus/pallets/parachain-system/src/lib.rs#L238) is used with the pallet. For some reason that `parameterized-consensus-hook` flag gets enabled on the integration tests, throwing a compilation error if we don't have this type in our runtimes.

Therefore I added `"cumulus-pallet-parachain-system/parameterized-consensus-hook"` to the `std` array for the runtimes. 

We really do not want this new type since we do not want to add any custom hook. My assumption is that this is not relevant for the runtimes on test, benchmarks or any `std` environment.

runtime/amplitude/src/lib.rs
```diff
@@ -1343,7 +1343,7 @@ impl farming::Config for Runtime {
 	type CurrencyId = CurrencyId;
 	type MultiCurrency = Currencies;
 	type ControlOrigin = EnsureRoot<AccountId>;
-	type WeightInfo = farming::weights::BifrostWeight<Runtime>;
+	type WeightInfo = ();
```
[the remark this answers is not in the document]
gianfra-t: Previously we were [using](https://github.com/pendulum-chain/bifrost/blob/71d2e5077de0c3b7e785dc2ea39b622300b7307b/pallets/farming/src/weights.rs#L65) the weights also from the pallet module. Now, as far as I understand, they [implement](https://github.com/bifrost-finance/bifrost/blob/polkadot-v1.1.0/pallets/farming/src/weights.rs#L80) the weights for the type `()` here, as they don't define anymore `BifrostWeight`.

This still is also generated with the same chain. Alternatively we could use the [ones](https://github.com/bifrost-finance/bifrost/blob/polkadot-v1.1.0/runtime/bifrost-kusama/src/weights/bifrost_farming.rs) defined for runtimes, but I don't see much difference in them.

runtime/amplitude/Cargo.toml
```diff
@@ -21,88 +21,89 @@ hex-literal = { version = "0.3.4", optional = true }
...
-orml-xcm-support = { git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.42" }
+orml-asset-registry = { git = "https://github.com/open-web3-stack/open-runtime-module-library.git", default-features = false, branch = "polkadot-v1.1.0" }
+orml-currencies = { git = "https://github.com/open-web3-stack/open-runtime-module-library.git", default-features = false, branch = "polkadot-v1.1.0" }
+orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library.git", default-features = false, branch = "polkadot-v1.1.0" }
+orml-tokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library.git", default-features = false, branch = "polkadot-v1.1.0" }
+orml-xtokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library.git", default-features = false, branch = "polkadot-v1.1.0" }
+orml-xcm-support = { git = "https://github.com/open-web3-stack/open-runtime-module-library.git", default-features = false, branch = "polkadot-v1.1.0" }
```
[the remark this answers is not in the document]
gianfra-t: Everything seems to compile without `.git` and it looks cleaner. I don't think there is any difference.

runtime/amplitude/src/xcm_config.rs
```diff
@@ -147,8 +148,8 @@ impl ShouldExecute for DenyReserveTransferToRelayChain {
 	fn should_execute<RuntimeCall>(
 		origin: &MultiLocation,
 		instructions: &mut [Instruction<RuntimeCall>],
-		_max_weight: XCMWeight,
-		_weight_credit: &mut XCMWeight,
+		_max_weight: Weight,
```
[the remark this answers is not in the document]
gianfra-t: For the struct in question `DenyReserveTransfer...` I think it is not needed in the logic (see [this one](https://github.com/paritytech/polkadot-sdk/blob/3c71db3efb7ed62dc0db950d3f090686803b10c5/polkadot/xcm/xcm-builder/src/barriers.rs#L461) defined in the sdk).

In fact, we might consider using the type directly from the sdk! But we need to analyze if it won't break anything.

runtime/integration-tests/src/amplitude_tests.rs
gianfra-t: Also, I tried using `11` but it did not work...

chain-extensions/token/src/lib.rs
```diff
@@ -228,7 +228,7 @@ where
 
 	<orml_currencies::Pallet<T> as MultiCurrency<T::AccountId>>::transfer(
 		currency_id,
-		env.ext().caller(),
+		env.ext().caller().account_id()?,
```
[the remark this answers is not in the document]
gianfra-t: I am not sure what we where doing before, but what error handling you have in mind? Right now the extension call would just fail.
About the likeliness, I have no idea how it could fail since it is something coming from the call itself.

runtime/integration-tests/src/genesis.rs
gianfra-t: Nope! I'll keep the l. 23 definition.

runtime/integration-tests/src/test_macros.rs
gianfra-t: I left it in case we want to explore the issue regarding the difference in fees I described in the description, of course before merging we can remove.

runtime/common/src/benchmarking/orml_asset_registry.rs
gianfra-t: I assumed we want to update to something different, `longest_vec` would not test the actual modification since it will return the same vec.

runtime/amplitude/src/lib.rs
gianfra-t: Yes! It does not necessarily mean that the storage migrations are to be coupled with an upgrade, but it will generally be the case. 

As soon as any runtime upgrade is executed on chain with this code declared, the migrations will happen, and this code becomes irrelevant to keep.

gianfra-t: In theory these are tested for idempotence, such that if we forget and upgrade many times with this code still there it will not affect the storage. But I think it is better to remove.

node/src/chain_spec.rs
gianfra-t: Oh! This must have been an automatic import, thanks for flagging.

pallets/parachain-staking/src/set.rs
gianfra-t: Alright! Let me try.

runtime/integration-tests/src/amplitude_tests.rs
b-yap: Don't change it. As of `release-polkadot-v1.1.0`, it's only up to version 5.

chain-extensions/token/src/lib.rs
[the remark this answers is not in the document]
ebma: What exactly is the return value of `account_id()`? I didn't find it in online documentation.

node/src/chain_spec.rs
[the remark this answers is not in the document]
ebma: I see. Yeah I guess it's fine to ignore, I was just curious where it comes from.

node/src/chain_spec.rs
```diff
@@ -617,6 +625,12 @@ fn amplitude_genesis(
 			batching_api: b"https://dia-00.pendulumchain.tech/currencies".to_vec(),
 			coin_infos_map: vec![],
 		},
+		token_allowance: Default::default(),
+		transaction_payment:  Default::default(),
+		treasury:  Default::default(),
+		treasury_buyout_extension: Default::default(),
+		vesting: Default::default(),
+		zenlink_protocol: Default::default(),
```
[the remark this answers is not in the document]
ebma: I found the [implementation of the construct_runtime ](https://paritytech.github.io/polkadot-sdk/master/src/frame_support_procedural/construct_runtime/mod.rs.html) because I wanted to see if the implicit declaration exposes everything by default. If I understand correctly, this is defined by the `tt_default_parts()` function that is automatically implemented on each pallet through the `#[pallet:pallet]` macro, see [here](https://paritytech.github.io/polkadot-sdk/master/src/frame_support_procedural/construct_runtime/mod.rs.html#35-36). I don't want to go further down the rabbit hole but this indicates that the implicit declaration might introduce unwanted side effects that are hard to debug (?). After all, the default genesis is [still available](https://github.com/pendulum-chain/pendulum/blob/7f9d96d1e0e7a7eebc46e034287bb24e733d93f5/pallets/orml-currencies-allowance-extension/src/lib.rs#L105-L123). @gianfra-t are you sure that the need for all these `Default::default()` statements arose after changing the definitions in the `construct_runtime` macro from explicit to implicit? Or could it be related to the version update?

If this really is required because we changed to the implicit definitions, then I wonder what else this entails that is not obvious and I'm a little scared that this gives us a headache later.

node/src/service.rs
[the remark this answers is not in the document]
ebma: I see, thanks 👍

runtime/integration-tests/src/amplitude_tests.rs
ebma: I see, thanks for pointing me to it 👍

ebma:
Thanks for explaining that part again 👍  Yes I fully agree that not having to patch all dependencies is the better approach, so let's stick to the runtimes from the `polkadot-sdk` repository. 

> About the weights, yes I didn't think about regenerating the weights again. Do you think it can be a bit too much for a single PR?

I would opt for including it in the same PR as the weights are only required due to the updated dependencies and we shouldn't roll out the runtime upgrade without regenerating them first. I would hope that this is not too much overhead.

chain-extensions/token/src/lib.rs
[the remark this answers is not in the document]
gianfra-t: It returns ` Result<&T::AccountId, DispatchError>`. Yeah, I had to go back to the IDE.

node/src/chain_spec.rs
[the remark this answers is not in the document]
gianfra-t: So I actually tried this. I rollback on one of the definitions, in this case `security` like so: 
```
Security: security::{Pallet, Call, Config<T>, Storage, Event<T>} = 67,
```
and I still get 

```
Security: security::{Pallet, Call, Config<T>, Storage, Event<T>} = 67,
```

This means that the explicit/implicit declaration does not really affect this new requirements for genesis. 
But, for the pallets that now need to add `Default()`, then declaring it explicitly without `Config` will indeed remove this requirement.

So, re-adding 
```
TokenAllowance: orml_currencies_allowance_extension::{Pallet, Storage, Call, Event<T>} = 80,
```
even forbids the use of declaring `token_allowance: Default::default(),` in genesis.

I don't know if this adds much information, at the end of the day, we still don't know exactly extra consequences could have leaving them implicit. I will try to find more info outside of reading the macro declaration.

Cargo.toml
b-yap: @gianfra-t  
I forgot to mention: `cargo clean` and `cargo update` didn't work for me because of: 
**`https://github.com/w3f//ring-vrf?rev=3ddc20?rev=3ddc20`**

`rev` part is called twice; probably because it is specified in the `git` field, followed by `rev` field.

I updated to 
```
[patch."https://github.com/w3f/ring-vrf"]
bandersnatch_vrfs = { git = "https://github.com/w3f//ring-vrf", version = "0.0.4", rev = "3ddc20" }
```
and the error disappeared.

node/src/chain_spec.rs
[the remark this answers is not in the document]
ebma: @TorstenStueber do you have any thoughts about changing from the explicit declaration to the implicit one?

Cargo.toml
gianfra-t: Thanks for finding this! I will change to the updated one.

node/src/chain_spec.rs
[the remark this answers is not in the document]
TorstenStueber: I used `cargo expand` to expand `construct_runtime` using the explicit (with `::{Pallet, Call, Storage, Event<T>}`) and the implicit definition for `pallet_vesting`. Then I diffed the result. The main difference (apart from a long list of definitions from serde) is that with the implicit definition the following type is defined:

```
pub type VestingConfig = pallet_vesting::GenesisConfig<Runtime>;
```

and the `pub struct RuntimeGenesisConfig` has an additional entry `pub vesting: VestingConfig`.

When using the explicit definition with `::{Pallet, Call, Config<T>, Storage, Event<T>}` (i.e. I added `Config<T>`) instead, then the result is equal to the implicit definition.

So I think the only difference in behavior between explicit and implicit definition affects all pallets where we did not explicitly state `Config<T>` before. I assume that this is the only side effect.

I think it is nicer to have a nice `construct_runtime` with implicit definitions and then live with stating all the `Default::default()` in the chain spec.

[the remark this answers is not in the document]
TorstenStueber: @ebma I used `cargo expand` to expand `construct_runtime` using the explicit (with `::{Pallet, Call, Storage, Event<T>}`) and the implicit definition for `pallet_vesting`. Then I diffed the result. The main difference (apart from a long list of definitions from serde) is that with the implicit definition the following type is defined:

```
pub type VestingConfig = pallet_vesting::GenesisConfig<Runtime>;
```

and the `pub struct RuntimeGenesisConfig` has an additional entry `pub vesting: VestingConfig`.

When using the explicit definition with `::{Pallet, Call, Config<T>, Storage, Event<T>}` (i.e. I added `Config<T>`) instead, then the result is equal to the implicit definition.

So I think the only difference in behavior between explicit and implicit definition affects all pallets where we did not explicitly state `Config<T>` before. I assume that this is the only side effect.

I think it is nicer to have a nice `construct_runtime` with implicit definitions and then live with stating all the `Default::default()` in the chain spec.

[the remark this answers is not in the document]
ebma: Cool, thanks for confirming @TorstenStueber 👍 I agree that the implicit version is more convenient and I'm also happy to switch to that. Let's keep it the way you changed it then @gianfra-t.

runtime/amplitude/src/lib.rs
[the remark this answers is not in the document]
ebma: I think it should be fine if we continue to leave it as is. So far, I'm only aware of the Zondax (ledger) team relying on it so I would assume that the other wallets don't necessarily check that number.

node/src/chain_spec.rs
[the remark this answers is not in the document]
gianfra-t: Good idea to test the expansion @TorstenStueber, didn't think of that. Let's keep the implicit.

gianfra-t:
I updated the weights by running the benchmarks again. Created a simple [script](.maintain/run_all_benchmarks.sh) for this that will run all the existing benchmarks on the weights folder, ignoring fails. @pendulum-chain/devs If you like this script we can keep it, I'll add it to the REAMDE.


Also added a fix for the failing `orlm-tokens-management` which only required to [increase](https://github.com/pendulum-chain/pendulum/pull/489/commits/1d49c521b42e33baf373ca50d6b81105ef700b27#diff-034d8e472a9794243adec40984451e87ce2e8d8360cb2fa07e828991064777c3R30) the balance of the tester account. I did it this way to avoid conversion errors (since in benchmarks is always hard to access values).

Only pallet `redeem` is not working for the `pendulum` runtime.
```
called `Result::unwrap()` on an `Err` value: Module(ModuleError { index: 53, error: [5, 0, 0, 0], message: Some("ExistentialDeposit") }) 
```
This is a bit surprising since also `amplitude` has the same existential deposit values, yet it works there.

We can either choose to investigate further, or use `amplitude`'s weights for this pallet, depending on priorities.

gianfra-t:
When you benchmark on `main` is harder to see, but running on this branch there is an `unwrap()` panicking, [this](https://github.com/pendulum-chain/spacewalk/blob/main/pallets/redeem/src/benchmarking.rs#L156) line precisely. 

So clearly the issue must be closer to the `VaultRegistry` config. But as you said, I don't understand how if the `ED` is the same. Do you recall any liquidation parameter that could be different between runtimes?

ebma:
I'm not 100% sure but maybe it's because of the differing values for the thresholds (e.g. [this](https://github.com/pendulum-chain/pendulum/blob/d186500fddba2958f60c106b4f76a1397bd74ab2/node/src/chain_spec.rs#L989)) of the vault registry in the genesis configuration of pendulum vs amplitude in `chain_spec.rs`. Can you try if changing all the thresholds to the same values as are used in Amplitude fixes the benchmark for Pendulum?

gianfra-t:
Nope, same error 🤔.

gianfra-t:
@ebma I pushed a [change](https://github.com/pendulum-chain/spacewalk/pull/551) to increase the amount of tokens issued before liquidating and this solves the fix. I suspect [this](https://github.com/pendulum-chain/spacewalk/blob/main/pallets/vault-registry/src/types.rs#L594) line and subsequent [transfer](https://github.com/pendulum-chain/spacewalk/blob/main/pallets/vault-registry/src/lib.rs#L1577), but I have no idea why it worked on Amplitude. In any case I executed both again with the same [parameter](https://github.com/pendulum-chain/spacewalk/pull/551/files#diff-72b7823392629ad279053970f76ca6eb8353619d749d7509e8a786f8f45835afR135).

Technically we should point to a new `rev` once the Spacewalk fix is merged.

@pendulum-chain/devs let me know if I missed an open question or remark since this PR grew a bit too much already!

ebma:
Although we don't 100% know which caused the issue, the amount used was very low so the fix is fine with me. Let's point to the new Spacewalk `rev` then.

chain-extensions/token/src/lib.rs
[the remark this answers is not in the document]
ebma: Okay, well if it's a `DispatchError` then I guess it's fine if we just keep the `?`.

runtime/amplitude/src/lib.rs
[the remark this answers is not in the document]
ebma: I see, makes sense to me.

b-yap (approved):

bogdanS98 (approved):
Looks good to me! Nothing else to add here 👍🏼

ebma (approved):

gianfra-t:
Sorry @b-yap, but if you can re-approve again please since @ebma and I where last to commit and now the review is stale.

ebma:
I'll just merge it for you @gianfra-t.

PR #493, opened 2024-09-23
Set version to `1.1.0` in all crates

### LINKED ISSUE #492: Set version that aligns to polkadot version in all crates
## Context

At the moment, there is no coherent versioning used in the pendulum crates. The `version` field was never updated for new releases. Going forward, the versions should be updated and align the polkadot dependency versions.

## Agreement

Align the node versions to the polkadot version, ie. for the v1.1.0 upgrade we would set the version to 1.1.0. This makes it easy to follow. If we want to update the collator binaries without changing the polkadot dependencies, we will add a -1 or -2, ... etc postfix to the version name. We set the same version on *all* crates in the pendulum repository. If needed, we can also add postfixes to the version numbers of other crates.

## TODO

Change the version of all crates to the current dependency version, ie. `1.1.0`.

ebma:
This sets the version of _all_ crates defined in the pendulum repository to `1.1.0`. This means, that some crates have major jumps or some versions are even set backward. But IMO it makes sense to have every crate defined with the same version and not just the `node` because every crate was updated to use the polkadot v1.1.0 dependencies so the version of every crate should reflect that.

Closes #492.

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pendulum-node"
-version = "0.1.0"
+version = "1.1.0"
 authors = ["Pendulum"]
 description = "The Pendulum/Amplitude collator node"
 homepage = "https://pendulumchain.org"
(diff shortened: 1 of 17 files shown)


node/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pendulum-node"
-version = "0.1.0"
+version = "1.1.0"
```
b-yap, line 3: This is noted. 

We'll just have to manually create a release note and tags of the Node for the time being.
:thumbsup:

pallets/orml-currencies-allowance-extension/Cargo.toml
```diff
@@ -26,27 +26,27 @@ pallet-balances = { git = "https://github.com/paritytech/polkadot-sdk", branch =
...
-  "sp-core/std",
-  "sp-std/std",
-  "sp-runtime/std",
-  "frame-support/std",
-  "frame-system/std",
-  "orml-currencies/std",
-  "orml-tokens/std",
-  "orml-traits/std",
-  "frame-benchmarking/std"
+    "serde/std",
```
b-yap, line 38: Is this caused by cargo fmt?

ebma: No, I think RustRover automatically formatted it like this 😬 A bit annoying that this is always changed back and forth but IIRC we already looked into an automatic formatting solution some time ago and nothing convenient was available. I think rustfmt doesn't support .toml files.

b-yap (approved):

PR #500, opened 2024-10-11
release: Bump pendulum spec version to 19

ebma:
Bumps the spec version of the pendulum runtime to 19. While this effectively makes us skip one version (18), we should still bump it to 19 to make it coherent as that's the version currently live on Amplitude.

Closing this will automatically trigger a release pipeline that builds the artifact and creates the github release, see [this](https://github.com/pendulum-chain/pendulum/blob/main/.github/workflows/release.yml).

Related to https://github.com/pendulum-chain/tasks/issues/329.


TorstenStueber (approved):
Let's go!!

PR #501, opened 2024-10-11
Use truly free balance (liquid) in chain extension

### LINKED ISSUE #483: Change chain extensions to return transferable balance
## Context
Related to this [ticket](https://github.com/pendulum-chain/portal/issues/485). Fixes to be made so that Nabla modals should show up as transferrable balance. 

Our chain extension exposes functions conforming to the ERC-20 interface. In ERC-20, the only function to get the balance of an address is using the `balanceOf()` call. At the moment, we are returning the 'free' balance of a user, see [here](https://github.com/pendulum-chain/pendulum/blob/991c7694084c22ee61716723d001163e746a1725/chain-extensions/token/src/lib.rs#L181). However, the `free_balance` is not necessarily the transferable balance, as it does not take into account the locked or reserved tokens. 

See the relation of the `system::account()` query to the transferable amount. 
image

image

## TODO

Change the `balance_of` function of the chain extension to return the transferable amount ie. `free - reserved - frozen`.

gianfra-t:
Closes #483 

Uses the `reducible_balance` fn from the `Inspect` trait. This is the truly liquid balance see [here](https://github.com/open-web3-stack/open-runtime-module-library/blob/polkadot-v1.1.0/tokens/src/lib.rs#L1808) of `orml_tokens` and [here](https://github.com/paritytech/polkadot-sdk/blob/release-polkadot-v1.1.0/substrate/frame/balances/src/impl_fungible.rs#L69) for `pallet_balances`.  Here frozen has a different meaning but the meaning of liquid is the same judging by how both use it in their `transfer_all` function.

There is no need to substract from `reserved` since these free and reserved are already mutually exclusive (see the reserve [extrinsic](https://github.com/open-web3-stack/open-runtime-module-library/blob/polkadot-v1.1.0/tokens/src/lib.rs#L1433)).

### Why not use MultiCurrency

Sadly, this trait does not implement the notion of a reducible balance nor of a frozen balance. It does implement this notion in `ensure_can_withdraw` but [this](https://github.com/open-web3-stack/open-runtime-module-library/blob/polkadot-v1.1.0/tokens/src/lib.rs#L736) won't actually tell us the amount, so the `Inspect` trait seems to be the only choice (that I could find). 
But this means we need to match between the native and token variants, which is not very ergonomic. Also, the return is encoded to avoid mismatches of `Balance` type.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15564,6 +15564,7 @@ dependencies = [
  "orml-currencies-allowance-extension",
  "orml-tokens",
  "orml-traits",
+ "pallet-balances",
  "pallet-contracts",
  "parity-scale-codec",
  "sp-core",
--- a/chain-extensions/token/Cargo.toml
+++ b/chain-extensions/token/Cargo.toml
@@ -21,6 +21,7 @@ sp-runtime = { git = "https://github.com/paritytech/polkadot-sdk", default-featu
 sp-std = { git = "https://github.com/paritytech/polkadot-sdk", branch = "release-polkadot-v1.1.0", default-features = false }
 sp-tracing = { git = "https://github.com/paritytech/polkadot-sdk", default-features = false, branch = "release-polkadot-v1.1.0" }
 sp-weights = { git = "https://github.com/paritytech/polkadot-sdk", default-features = false, branch = "release-polkadot-v1.1.0" }
+pallet-balances = { git = "https://github.com/paritytech/polkadot-sdk", default-features = false, branch = "release-polkadot-v1.1.0" }
 pallet-contracts = { git = "https://github.com/paritytech/polkadot-sdk", default-features = false, branch = "release-polkadot-v1.1.0" }
 
 # Open Runtime Module Library
@@ -56,4 +57,5 @@ std = [
     "spacewalk-primitives/std",
     "chain-extension-common/std",
     "sp-std/std",
+    "pallet-balances/std"
 ]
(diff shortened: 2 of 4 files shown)


ebma:
@gianfra-t is this ready for review? You didn't add reviewers.

gianfra-t:
Yes @ebma it is, but tests are failing (I was waiting for them). I will check if it is related.

gianfra-t:
I will ran format afterwards since it will as always complicate the review of actual changes.

ebma (approved):
Interesting, I didn't know about the `reducible_balance` function. Makes sense to use it but unfortunate that it's not implemented on the `MultiCurrency` trait. 

Looks good to me overall 👍

TorstenStueber:
Is the failing CI due to clippy? Isn't it the Test job that fails with the following error?

```
Oct 14 13:51:18.446 ERROR parachain_staking::pallet: 💥 keeping old session because of empty collator set!
```

gianfra-t:
But that has been a very common log we see almost always we start the chain in tests. I doubt it will make the CI fail.

It almost seems like it didn't run the last 3 jobs.

gianfra-t:
Ohhh I checked the logs of the CI again and I see:

```
System.IO.IOException: No space left on device : '/home/runner/runners/2.320.0/_diag/Worker_20241014-124125-utc.log
```

So that's it.

PR #502, opened 2024-10-17
Amend node version up

ebma:
Follow-up to https://github.com/pendulum-chain/pendulum/pull/485, fixing a minor issue.

After merging #485 I tried to run the [workflow](https://github.com/pendulum-chain/pendulum/actions/runs/11380853257) but it failed because the file name was wrong. 

- [x] Fix reference to file
- [x] Remove unused files


gianfra-t (approved):
👍. Will we never use the runtime version up script again?

b-yap:
Looks fine. Merged.

PR #510, opened 2024-11-27
Update to Polkadot SDK 1.6.0 

### LINKED ISSUE #505: Update to Polkadot-sdk 1.6.0 - Pendulum
Related to master ticket [tasks/#419](https://github.com/pendulum-chain/tasks/issues/419)

We need to perform the upgrade of Pendulum to [polkadot-sdk release v1.6.0](https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.6.0).

This involves updating all corresponding dependencies in this repository:
- Self-maintained pallets
- Chain extension code
- Runtimes (Including storage migrations)
- Node client

**Before performing the runtime upgrade** with this changes it is essential to have [checked]() which are the necessary migrations and added the migration code, if needed.

ebma:
## TODO
- [x] Update the declared dependency versions (mostly done). `cargo check` does not work yet. 
- [x] Add code changes to adjust to breaking changes
- [x] (Optional) Group everything in a global workspace file. I wanted to do this last so we can first focus on making everything work in general before dealing with the errors that potentially arise only due to mistakes with the migration to the global workspace. 
- [x] Run `try-runtime` CLI tests, to verify the storage version updates.
- [ ] Verify new configuration parameters chosen for pallets, necessary after the upgrade.
- [x] Replace the forked repositories (bifrost, zenlink) that avoid using crates.io with forks from `pendulum-chain`.
## Important Note
Package `runtime-integration-tests` will not be updated on this PR, as it requires further dependency refactor that potentially consist of moving all polkadot dependencies to it's `crates-io` definition (mostly due to the removal of `asset-hub` runtimes from `polkadot-sdk`).

Closes #505.


runtime/pendulum/src/lib.rs
```diff
@@ -468,8 +475,9 @@ impl pallet_balances::Config for Runtime {
 	type ReserveIdentifier = ReserveIdentifier;
 	type FreezeIdentifier = ();
 	type MaxFreezes = ();
-	type MaxHolds = ConstU32<1>;
+	type MaxHolds = ConstU32<2>;
```
gianfra-t, line 478: Otherwise, we get an error `MaxHolds should be greater than or equal to the number of hold reasons: 1 < 2`, which is an internal check of the pallet.
Apparently the `RuntimeHoldReason` has 2 variants.

runtime/pendulum/src/lib.rs
```diff
@@ -719,6 +745,7 @@ parameter_types! {
 	pub const Burn: Permill = Permill::from_percent(0);
 	pub const TreasuryPalletId: PalletId = PalletId(*b"py/trsry");
 	pub const MaxApprovals: u32 = 100;
+	pub const PayoutSpendPeriod: BlockNumber = 30 * DAYS;
```
gianfra-t, line 748: I selected this arbitrarily. We might want to change.

runtime/pendulum/src/lib.rs
```diff
@@ -1005,6 +1042,8 @@ parameter_types! {
 	pub const MaxSubAccounts: u32 = 100;
 	pub const MaxAdditionalFields: u32 = 100;
 	pub const MaxRegistrars: u32 = 20;
+	pub const ByteDeposit: Balance = MILLIUNIT;
```
gianfra-t, line 1045: Also arbitrary and we may tweak to our business req.

runtime/pendulum/src/lib.rs
```diff
@@ -527,20 +538,29 @@ impl cumulus_pallet_aura_ext::Config for Runtime {}
...
+
+parameter_types! {
+	pub const RelayOrigin: AggregateMessageOrigin = AggregateMessageOrigin::Parent;
 }
 
 impl cumulus_pallet_dmp_queue::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
-	type XcmExecutor = XcmExecutor<XcmConfig>;
-	type ExecuteOverweightOrigin = EnsureRoot<AccountId>;
+	type DmpSink = frame_support::traits::EnqueueWithOrigin<MessageQueue, RelayOrigin>;
```
gianfra-t, line 562: The pallet is deprecated, but we keep it to until it drains it's messages (to the replacement), assuming there are some after the update.

I couldn't really find a lot of information on this replacement.

runtime/pendulum/src/lib.rs
```diff
@@ -1016,15 +1055,49 @@ impl pallet_identity::Config for Runtime {
...
+	type MaxUsernameLength = ConstU32<32>;
+}
+
+parameter_types! {
+    pub MessageQueueServiceWeight: Weight =
+        Perbill::from_percent(25) * RuntimeBlockWeights::get().max_block;
+}
+
+use parachains_common::message_queue::NarrowOriginToSibling;
+impl pallet_message_queue::Config for Runtime {
```
gianfra-t, line 1081: This general queue pallet can work as dmp, xcmp, etc. We use it now as a replacement for the deprecated `dmp_queue` pallet.

runtime/pendulum/src/weights/parachain_staking.rs
```diff
@@ -24,7 +24,7 @@
 // --repeat
 // 5
```
ebma, line 25: Maybe we should adhere to the 'best practices' used in the [docs for benchmarking](https://docs.substrate.io/test/benchmark/#running-benchmarks) and have it repeat not `5` but `20` times.

gianfra-t: Yes I changed it because it takes ages with `20` times. It already takes like 30 minutes with 5. And then I remembered that was the reason we disabled it from the script to run all benchmarks.

ebma: Ah okay 😅 fair enough.

node/src/chain_spec.rs
```diff
@@ -857,24 +763,22 @@ fn foucoco_genesis(
 		treasury_buyout_extension: Default::default(),
 		vesting: Default::default(),
 		zenlink_protocol: Default::default(),
-	}
+	};
+
+	serde_json::to_value(genesis_config).expect("Serialization of genesis config should work")
 }
 
+const MAX_SAFE_INTEGER_JSON: u128 = 1 << 53;
```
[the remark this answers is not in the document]
gianfra-t: To the top it goes!

runtime/amplitude/src/lib.rs
```diff
@@ -527,20 +536,29 @@ impl cumulus_pallet_aura_ext::Config for Runtime {}
...
+parameter_types! {
+	pub const RelayOrigin: AggregateMessageOrigin = AggregateMessageOrigin::Parent;
 }
 
 impl cumulus_pallet_dmp_queue::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
-	type XcmExecutor = XcmExecutor<XcmConfig>;
-	type ExecuteOverweightOrigin = EnsureRoot<AccountId>;
+	type DmpSink = frame_support::traits::EnqueueWithOrigin<MessageQueue, RelayOrigin>;
+	type WeightInfo = ();
```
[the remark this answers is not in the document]
gianfra-t: The defaults weights are [implemented](https://github.com/paritytech/polkadot-sdk/blob/release-polkadot-v1.6.0/cumulus/pallets/dmp-queue/src/weights.rs#L141) for `()` so it's essentially the same. You think we should add it to the pallet we benchmark in our chain?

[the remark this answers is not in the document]
ebma: No need to benchmark them extra probably. I can see that e.g. Astar are also using default weights for similar pallets, see [here](https://github.com/AstarNetwork/Astar/blob/765d44ba1e9210279088ff765aff5ee6979601d2/runtime/shiden/src/lib.rs#L317). However, it might be nice to make the assignment explicit so it's more obvious that there are weights. Not urgent, just a preference on my side.

runtime/pendulum/src/lib.rs
```diff
@@ -748,6 +775,14 @@ impl pallet_treasury::Config for Runtime {
 	type WeightInfo = pallet_treasury::weights::SubstrateWeight<Runtime>;
 	type MaxApprovals = MaxApprovals;
 	type SpendOrigin = frame_support::traits::NeverEnsureOrigin<u128>;
+	type Beneficiary = AccountId;
+	type BeneficiaryLookup = IdentityLookup<Self::Beneficiary>;
+	type Paymaster = PayFromAccount<Balances, PendulumTreasuryAccount>;
```
ebma, line 780: I think we should rather use [PayAssetFromAccount](https://github.com/paritytech/polkadot-sdk/blob/77c78e1561bbe5ee0ecf414312bae82396ae6d11/substrate/frame/support/src/traits/tokens/pay.rs#L118) as is done [here](https://github.com/paritytech/polkadot-sdk/blob/d822e07d51dda41982291dc6582a8c4a34821e94/substrate/bin/node/runtime/src/lib.rs#L1306). 
IIRC the fees of our XCM transfers are charged in the transferred asset, so the treasury will have more than just the native token. With `PayFromAccount` I think we'd only be able to transfer native tokens out of it with treasury proposals.

gianfra-t: Makes sense! Thanks for finding it.

runtime/pendulum/src/lib.rs
```diff
@@ -947,6 +982,7 @@ impl pallet_vesting::Config for Runtime {
 	type UnvestedFundsAllowedWithdrawReasons = UnvestedFundsAllowedWithdrawReasons;
 	type WeightInfo = pallet_vesting::weights::SubstrateWeight<Runtime>;
 	const MAX_VESTING_SCHEDULES: u32 = 10;
+	type BlockNumberProvider = runtime_common::RelayChainBlockNumberProvider<Runtime>;
```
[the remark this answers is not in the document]
gianfra-t: Don't remember at this stage, but if it allows us to get rid of `RelayChainBlockNumberProvider` then fine by me!

runtime/pendulum/src/lib.rs
```diff
@@ -1016,15 +1055,49 @@ impl pallet_identity::Config for Runtime {
...
+	type SigningPublicKey = <Signature as sp_runtime::traits::Verify>::Signer;
+	type UsernameAuthorityOrigin = EnsureRoot<Self::AccountId>;
+	type PendingUsernameExpiration = PendingUsernameExpiration;
+	type MaxSuffixLength = ConstU32<8>;
+	type MaxUsernameLength = ConstU32<32>;
+}
+
+parameter_types! {
+    pub MessageQueueServiceWeight: Weight =
+        Perbill::from_percent(25) * RuntimeBlockWeights::get().max_block;
```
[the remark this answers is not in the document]
gianfra-t: Probably because they also defined [here](https://github.com/paritytech/polkadot-sdk/blob/d822e07d51dda41982291dc6582a8c4a34821e94/substrate/bin/node/runtime/src/lib.rs#L487C6-L487C28) the weight of the scheduler, and it must be consistent. Since we also define `80%` we must choose `20%` then.

runtime/foucoco/src/lib.rs
```diff
@@ -1491,6 +1566,9 @@ construct_runtime!(
 		// Asset Metadata
 		AssetRegistry: orml_asset_registry = 91,
 
+		// Addition pallet message queue
+		MessageQueue: pallet_message_queue = 92,
```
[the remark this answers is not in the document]
gianfra-t: Actually, for Foucoco since we already made the upgrade, we probably cannot change it right?

runtime/amplitude/src/lib.rs
```diff
@@ -1488,7 +1561,9 @@ construct_runtime!(
 		// Asset Metadata
 		AssetRegistry: orml_asset_registry = 91,
 
-		VestingManager: vesting_manager = 100
+		VestingManager: vesting_manager = 100,
+
+		MessageQueue: pallet_message_queue = 222, // TODO define a convention for these type of additions.
```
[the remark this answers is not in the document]
gianfra-t: Yup! I must have added the numbers in different days. Set to `110`. Probably `10` is more than fine to group new pallets with similar logic.

runtime/foucoco/src/lib.rs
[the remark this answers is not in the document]
gianfra-t: Changed to 110.

runtime/amplitude/src/lib.rs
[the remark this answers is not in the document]
gianfra-t: Changed!

ebma (commented):
Looks good to me, let's finally get this out of the way.

PR #517, opened 2025-02-03
Adds config to enable warp sync.

gianfra-t:
Adds `warp_sync_params` config. Based on cumulus client [config](https://github.com/paritytech/cumulus/blob/574f425fa2b9c822a1417af705169ab3a1835e24/client/service/src/lib.rs#L344-L368).

Standalone chain will not need this config enabled, therefore we can safely avoid chain id and relay interface.

### DIFF
--- a/node/src/service.rs
+++ b/node/src/service.rs
@@ -2,7 +2,7 @@
 
 // std
 use std::{sync::Arc, time::Duration};
-
+use codec::Decode;
 use cumulus_client_cli::{CollatorOptions, RelayChainMode};
 // Local Runtime Types
 use runtime_common::{opaque::Block, AccountId, Balance, Index as Nonce};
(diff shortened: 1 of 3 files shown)


ebma (approved):
Looks good to me 👍

gianfra-t:
I'll try to get the test working, at least locally. We're having the storage issue again.

node/src/service.rs
```diff
@@ -100,6 +99,8 @@ impl ParachainRuntimeApiImpl for amplitude_runtime::RuntimeApiImpl<Block, Amplit
 impl ParachainRuntimeApiImpl for pendulum_runtime::RuntimeApiImpl<Block, PendulumClient> {}
 impl ParachainRuntimeApiImpl for foucoco_runtime::RuntimeApiImpl<Block, FoucocoClient> {}
 
+const LOG_TARGET_SYNC: &str = "sync::cumulus";
```
ebma, line 102: Ah just one thing. Let's maybe remove this constant and use `"sync::cumulus"` in the places where the constant is mentioned. To me, having this constant gives more significance to this log target level although it's actually not really important. And for the other log with target `"cumulus-network"` we also didn't define a constant so let's just remove this. I know it was like this in the template.

ebma (approved):

Issue #518, opened 2025-02-04
Parachain stopped finalizing blocks after upgrade


gflow33:
I am running a pendulum rpc node and after upgrading from https://github.com/pendulum-chain/pendulum/releases/tag/1.1.0 to https://github.com/pendulum-chain/pendulum/releases/tag/1.6.0 the node stopped finalizing blocks and produced the errors in the log below

The container in which it runs is on ubuntu 22.04 and run with the following flags

```
./pendulum-collator \
--name="\U0001F6E1 DWELLIR PENDULUM RPC \U0001F6E1" \
--chain=pendulum \
--pruning=archive \
--rpc-max-connections=1000 \
--rpc-external --rpc-port=9933 \
--rpc-cors=all \
--rpc-methods=unsafe \
--prometheus-external \
--runtime-cache-size=16 \
--max-runtime-instances=32 \
--execution wasm \
-- \
--chain=polkadot
```

Node logs 

```
Feb 04 16:11:15 juju-ab466c-0 polkadot[2284297]: 2025-02-04 16:11:15 [Parachain] Evicting failed runtime instance error=Runtime panicked: Could not lookup information required to validate the transaction
Feb 04 16:11:15 juju-ab466c-0 polkadot[2284297]: 2025-02-04 16:11:15 [Parachain] Block prepare storage changes error: Error at calling runtime api: Execution failed: Runtime panicked: Could not lookup information required to validate the transaction
Feb 04 16:11:15 juju-ab466c-0 polkadot[2284297]: 2025-02-04 16:11:15 [Parachain] 💔 Error importing block 0x7f9123dfff8be6108cd59a5b3346dbcbb6e1f0a54f5c0bde7c140db14f9d2466: consensus error: Import failed: Error at calling runtime api: Execution failed: Runtime panicked: Could not lookup information required to validate the transaction
Feb 04 16:11:15 juju-ab466c-0 polkadot[2284297]: ====================
Feb 04 16:11:15 juju-ab466c-0 polkadot[2284297]: Version: 1.6.0-93c9dbe0542
Feb 04 16:11:15 juju-ab466c-0 polkadot[2284297]:    0: sp_panic_handler::set::{{closure}}
Feb 04 16:11:15 juju-ab466c-0 polkadot[2284297]:    1: std::panicking::rust_panic_with_hook
Feb 04 16:11:15 juju-ab466c-0 polkadot[2284297]:    2: std::panicking::begin_panic_handler::{{closure}}
[...]

PR #521, opened 2025-02-28
Replace native execution.

gianfra-t:
Fixes issue [518](https://github.com/pendulum-chain/pendulum/issues/518).

### Context

Since the metadata-hash chain extension [cannot be used](https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_docs/guides/enable_metadata_hash/index.html) with the native runtime execution, and it's use seems to be [deprecated](https://github.com/paritytech/substrate/pull/14387) as well, we replace the use of `NativeElseWasmExecutor` and use `WasmExecutor` directly.


node/src/service.rs
```diff
@@ -188,16 +151,14 @@ where
 		.default_heap_pages
 		.map_or(DEFAULT_HEAP_ALLOC_STRATEGY, |h| HeapAllocStrategy::Static { extra_pages: h as _ });
 
-	let wasm = WasmExecutor::builder()
+	let executor = ParachainExecutor::builder()
 		.with_execution_method(config.wasm_method)
```
ebma, line 155: Isn't the `execution_method` field deprecated anyway? Or is it just the CLI field? I remember seeing warnings when starting the node.

node/src/service.rs
```diff
@@ -53,33 +51,38 @@ use sc_client_api::Backend;
...
+pub type ParachainHostFunctions = (
+	frame_benchmarking::benchmarking::HostFunctions,
+);
+#[cfg(not(feature = "runtime-benchmarks"))]
+pub type ParachainHostFunctions = (
+	sp_io::SubstrateHostFunctions,
+);
 
-type ParachainBlockImport<RuntimeApi, Executor> = TParachainBlockImport<
+pub type ParachainExecutor = WasmExecutor<ExtendedHostFunctions<sp_io::SubstrateHostFunctions, ParachainHostFunctions>>;
```
ebma, line 69: Nit: You now define `sp_io::SubstrateHostFunctions` twice in the `not runtime-benchmarks` case. Let's maybe do it like [Centrifuge](https://github.com/centrifuge/centrifuge-chain/blob/6ca83c7646dc9cc0c8b7d267c649904505f639fc/node/src/service.rs#L43-L52) does and define it in a tuple. 

```
#[cfg(feature = "runtime-benchmarks")]
type ParachainHostFunctions = (
	sp_io::SubstrateHostFunctions,
	frame_benchmarking::benchmarking::HostFunctions,
);

#[cfg(not(feature = "runtime-benchmarks"))]
type ParachainHostFunctions = sp_io::SubstrateHostFunctions;

pub type ParachainExecutor = WasmExecutor<ParachainHostFunctions>;
```

ebma (approved):
So you managed to sync the chain on Foucoco with this change? 🥹 Great find 🙏

node/src/service.rs
gianfra-t: Ah, good call. We can remove this and get rid of the warning.

node/src/service.rs
gianfra-t: Like it.

node/src/service.rs
gianfra-t: I checked this method and it seems that it is not for selecting Native/wasm, but rather _how_ to [instantiate](https://github.com/paritytech/polkadot-sdk/blob/3798ff7f5464348ec3cd05798081d4cc01728500/substrate/client/executor/src/wasm_runtime.rs#L49) the wasm. I think we should keep it.

But you're right that there is a warning for the `--execution` flag, which now makes even less sense to have. I'll try to find a way to remove it.

ebma: Okay understood.

ebma (approved):

ebma:
@gianfra-t shall we merge this and create a new release? We can roll out and test it on Foucoco afterward.

prayagd:
Should we close the ticket of this PR?

ebma:
Not yet, We have to check how it behaves on Foucoco after rolling it out.

PR #541, opened 2025-09-10
Transfer staking rewards from the treasury.

gianfra-t:
Replace mint operation when claiming rewards with a transfer from the treasury account.


gianfra-t:
@ebma pinging for re-approve, all tests seem to be okay now.

